PR #55, opened 2019-09-13
:tada: Support double and multi click events on gui elements

Flatken:


Schneegans:
Is it correct, that the `SendMouseClickEvent` method is called twice for double clicks?

PR #83, opened 2019-11-14
:beetle: Fix two small GL bugs

bernstein:
Unbind pbo after usage in LuminanceMipMap constructor and fix type of a uniform variable in the gltf fragment shader.


Schneegans:
Thank you!

Issue #90, opened 2019-12-10
Tests


TheJhonny007:
Add unit testing to CosmoScout VR.

Things to consider:
- Which framework to use?
  - Google Test
  - Boost.Test
  - Catch
  - DocTest
  - ...
- What scope?
  - Unit Tests?
  - Integration Tests?
- Make runs mandatory or optional.
- Include into GitHub Actions?
- How to integrate into the project
  - Test suit per module?
  - What about plugins?

Schneegans:
I vote for DocTest. There is also an example for running test from shared objects and plugins: https://github.com/onqtam/doctest/tree/master/examples/executable_dll_and_plugin

We could write the tests in a dedicated `tests.cpp` for each module / plugin. Then a flag `--run-tests` of the `cosmoscout` executable could run all tests. This could be available in debug mode only.

PR #92, opened 2019-12-16
Feature/doctest test

TheJhonny007:
Solves issue #90


Schneegans:
Shall I look into this?

TheJhonny007:
> Shall I look into this?

Would be cool, since you broke it :yum: 
It's probably the same reason, why the CI Windows build fails.

Issue #86, opened 2019-12-02
IDE setup documentation


TheJhonny007:
Document the project setup with different IDEs.

- [ ] Visual Studio
- [ ] Visual Studio Code
- [ ] Sublime
- [ ] CLion
- [ ] Eclipse?
- [ ] CodeBlocks?

PR #94, opened 2019-12-16
Feature/clion docs

TheJhonny007:
Solves issue #86 with the most used IDEs for now.


TheJhonny007:
Somehow a dependency from another branch got in here, which fails the Linux build... I will fix that later.

PR #95, opened 2019-12-19
Rework CosmoScout GUI

octfx:
This pull request consolidates all main UI elements into one file. Furthermore all js functions are bundled in well defined api classes.

A global `CosmoScout` object acts as the main api entry point. Registered apis are callable via `CosmoScout.call('api', 'method', ...args)` or `CosmoScout.api.method(...args)`.

The main UI is contained in `cosmoscout.html`. This file also holds all addable html fragments in `<template id="example-template">` tags. Template contents can be loaded via `CosmoScout.loadTemplateContent(id)`, where `id` is the templates id without `-template` suffix.

Related plugin pull requests:
- [x] https://github.com/cosmoscout/csp-sharad/pull/3
- [x] https://github.com/cosmoscout/csp-fly-to-locations/pull/3
- [x] https://github.com/cosmoscout/csp-measurement-tools/pull/3
- [x] https://github.com/cosmoscout/csp-lod-bodies/pull/8
- [x] https://github.com/cosmoscout/csp-stars/pull/3
- [x] https://github.com/cosmoscout/csp-atmospheres/pull/3
- [x] https://github.com/cosmoscout/csp-trajectories/pull/3
- [x] https://github.com/cosmoscout/csp-anchor-labels/pull/2


TheJhonny007:
Resizing the window after the GUI has loaded results in a crash. The crash is the result of illegal memory access here: https://github.com/cosmoscout/cosmoscout-vr/blob/feature/cosmoscout-gui/src/cs-gui/internal/RenderHandler.cpp#L42

Copying the buffer from the `OnPaint()` method beforehand resolves this issue, but comes at the cost of horrible framerates.

TheJhonny007:
> Resizing the window after the GUI has loaded results in a crash. The crash is the result of illegal memory access here: https://github.com/cosmoscout/cosmoscout-vr/blob/feature/cosmoscout-gui/src/cs-gui/internal/RenderHandler.cpp#L42
> 
> Copying the buffer from the `OnPaint()` method beforehand resolves this issue, but comes at the cost of horrible framerates.

This should be fixed now. It takes around one millisecond to copy the framebuffer on my machine with a resolution of 5120x1440 pixels. Maybe there is an even better method?

Schneegans:
Thank you very much! I will look into this as soon as I find the time. In the meantime, you could create all required pull requests in the submodules and link them here.

Schneegans:
After a very quick look I found one issue: Why did you move plugin code to the sidebar api? For example, `addMeasurementTool()` should definitely be in `csp-measurement-tools` and not in CosmoScout core...

octfx:
> After a very quick look I found one issue: Why did you move plugin code to the sidebar api? For example, `addMeasurementTool()` should definitely be in `csp-measurement-tools` and not in CosmoScout core...

This is indeed an error. I'll move the corresponding code back into the plugins repository.

resources/gui/js/api.js
```diff
@@ -0,0 +1,543 @@
...
+  /**
+   * @see {initInputs}
+   * @see {callNative}
+   * Adds an onclick listener to every element containing [data-call="methodname"]
+   * The method name gets passed to CosmoScout.callNative.
+   * Arguments can be passed by separating the content with ','
+   * E.g.: fly_to,Africa -> CosmoScout.callNative('fly_to', 'Africa')
+   *       method,arg1,...,argN -> CosmoScout.callNative('method', arg1, ..., argN)
+   */
+  static initDataCalls() {
```
Schneegans: I think this method can be improved somehow. Here are some issues I see:
* Does it work for multiple number arguments?
* What about boolean arguments?
* Last but not least, `data-call="set_quality,5f"` looks a bit ugly

Have you considered just to use `eval()`? It should be possible to write `data-call="CosmoScout.callNative('set_quality', 5)"` and on click you just call `eval(input.dataset.call)`. What do you think?

resources/gui/cosmoscout.html
```diff
@@ -0,0 +1,546 @@
...
+            <a class="btn light-glass"
+               data-toggle="tooltip-bottom"
+               title="Land on surface"
+               data-call="navigate_to_surface">
+                <i class="material-icons">flight_land</i>
+            </a>
+        </div>
+    </div>
+
+    <div id="left-side-bar" class="side-bar left">
```
Schneegans: For consistency, you could remove the .side-bar and .left classes here. ` <div id="side-bar">` should be sufficient; we will not have a right sidebar anytime soon.

resources/gui/cosmoscout.html
```diff
@@ -0,0 +1,546 @@
...
+                <div id="progress-bar">
+                </div>
+            </div>
+            <div id="lower">loading core engine ...</div>
+        </div>
+        <div id="version">unknown version</div>
+    </div>
+</div>
+
+<div id="cosmo-scout">
```
Schneegans: Usually, `cosmoscout` is written without a dash :smiley:.

resources/gui/js/api.js
```diff
@@ -0,0 +1,543 @@
...
+  /**
+   * Global entry point to call any method on any registered api
+   *
+   * @param api {string} Name of api
+   * @param method {string} Method name
+   * @param args {string|number|boolean|Function|Object} Arguments to pass through
+   * @return {*}
+   */
+  static call(api, method, ...args) {
+    if (method !== 'setUserPosition' && method !== 'setNorthDirection' && method !== 'setSpeed' && method !== 'setDate') {
```
Schneegans: This can be removed, I guess :smiley:.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+    $(`#${id}`).selectpicker('render');
+  }
+
+  /**
+   * Sets an elevation data copyright tooltip
+   * TODO Remove jQuery
+   *
+   * @param copyright {string}
+   */
+  setElevationDataCopyright(copyright) {
```
Schneegans: This should be in `csp-lod-bodies`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+    $('#img-data-copyright').tooltip({ title: `© ${copyright}`, placement: 'top' });
+  }
+
+  /**
+   * Sets a map data copyright tooltip
+   * TODO Remove jQuery
+   *
+   * @param copyright {string}
+   */
+  setMapDataCopyright(copyright) {
```
Schneegans: This should be in `csp-lod-bodies`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+
+    this._settings.appendChild(tab);
+  }
+
+  /**
+   * Passthrough call to CosmoScout.clearHtml
+   *
+   * @param id {string}
+   */
+  clearContainer(id) {
```
Schneegans: This should be removed, I guess. All usages of this should be replaced by `CosmoScout.clearHtml(id);`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+    CosmoScout.clearHtml(id);
+  }
+
+  /**
+   * TODO UNUSED / Remove jQuery
+   *
+   * @param id {string}
+   */
+  clearDropdown(id) {
```
Schneegans: This could be moved to the `CosmoScout` object. Somewhere close to `initSlider()`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Adds an option to a dropdown
+   * TODO remove jQuery
+   *
+   * @param id {string} DropDown ID
+   * @param value {string|number} Option value
+   * @param text {string} Option text
+   * @param selected {boolean} Selected flag
+   */
+  addDropdownValue(id, value, text, selected = false) {
```
Schneegans: This could be moved to the `CosmoScout` object. Somewhere close to `initSlider()`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  /**
+   * TODO UNUSED
+   *
+   * @param id {string}
+   * @param value {string|number}
+   */
+  setDropdownValue(id, value) {
```
Schneegans: This could be moved to the `CosmoScout` object. Somewhere close to `initSlider()`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+    $(`#${id}`).selectpicker('val', value);
+  }
+
+  /**
+   * TODO UNUSED
+   * @param id {string}
+   */
+  setRadioChecked(id) {
```
Schneegans: This could be moved to the `CosmoScout` object. Somewhere close to `initSlider()`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+    this.setCheckboxValue(id, true);
+  }
+
+  /**
+   * TODO UNUSED
+   * @param id {string}
+   * @param value {boolean}
+   */
+  setCheckboxValue(id, value) {
```
Schneegans: This could be moved to the `CosmoScout` object. Somewhere close to `initSlider()`.

resources/gui/js/apis/sidebar.js
```diff
@@ -0,0 +1,194 @@
...
+      element.checked = value === true;
+    }
+  }
+
+  /**
+   * TODO UNUSED
+   * @param id {string}
+   * @param value {string}
+   */
+  setTextboxValue(id, value) {
```
Schneegans: This could be moved to the `CosmoScout` object. Somewhere close to `initSlider()`.

src/cosmoscout/Application.cpp
```diff
@@ -400,6 +397,7 @@ void Application::FrameUpdate() {
 
     // Hide the loading screen after several frames.
     if (GetFrameCount() == mHideLoadingScreenAtFrame) {
+      mGuiManager->getGui()->callJavascript("CosmoScout.initInputs");
```
Schneegans: Why do we require this call twice?

src/cosmoscout/Application.cpp
```diff
@@ -510,10 +508,10 @@ void Application::FrameUpdate() {
       double heightDiff    = polar.z / mGraphicsEngine->pHeightScale.get() - surfaceHeight;
 
       if (!std::isnan(polar.x) && !std::isnan(polar.y) && !std::isnan(heightDiff)) {
-        mGuiManager->getStatusBar()->callJavascript("set_user_position",
+        mGuiManager->getGui()->callJavascript("CosmoScout.call", "timeline", "setUserPosition",
```
Schneegans: I find `mGuiManager->getGui()->callJavascript("CosmoScout.timeline.setNorthDirection", angle);` much more readable. Should  we replace all `callJavascript`s?

src/cs-core/GuiManager.hpp
```diff
@@ -123,6 +123,12 @@ class CS_CORE_EXPORT GuiManager {
   /// @param jsFile The javascript file that contains the source code.
   void addScriptToSideBarFromJS(std::string const& jsFile);
```
Schneegans: These two methods should be called `addScriptToGui` and `addScriptToGuiFromJS`.

Schneegans:
There is a feature checkbox called **Clear API between C++ and JavaScript** in the main README.md. You can now definitely check this together with **Flexible User Interface** 😃.

resources/gui/js/api.js
octfx: This is now resolved by using eval. The contents of the data-call attribute will be passed to `CosmoScout.callNative`.  

Example:
```
<a data-call="'set_quality',5"> => CosmoScout.callNative('set_quality', 5)
```
Notice `'` around `set_quality`. This is needed to denote strings, as `eval` wouldn't parse it as one otherwise.

resources/gui/cosmoscout.html
octfx: Updated. For consistency `side-bar` got renamed to `sidebar`.

resources/gui/js/api.js
octfx: Fully removed `call` and replaced it with direct object accesses.

resources/gui/js/apis/sidebar.js
octfx: Moved all methods into the new `LODBodyApi`.

resources/gui/js/apis/sidebar.js
octfx: Moved all methods into the main `CosmoScout` api.

src/cosmoscout/Application.cpp
octfx: This was indeed a wrongful double call. Removed the call from `cosmoscout.html`.  
The one from `Application.cpp` should be the 'right' call, as at this point in time all needed html is present in the gui.

src/cosmoscout/Application.cpp
octfx: Renamed all calls to `CosmoScout.<api>.<method>`.

src/cs-core/GuiManager.hpp
octfx: Renamed.

Schneegans:
> 
> 
> Todo:
> 
>     * [ ]  `fly_to` is not working
>       
>       * Neither the timeline event for Apollo 17 nor calling it with the name of a celestial body

Is this still an issue?

octfx:
> > Todo:
> > ```
> > * [ ]  `fly_to` is not working
> >   
> >   * Neither the timeline event for Apollo 17 nor calling it with the name of a celestial body
> > ```
> 
> Is this still an issue?

I'll have a look tomorrow. Last I checked the bug seemed to be not GUI/js related.

TheJhonny007:
> Thank you so much! These changes significantly improve large parts of CosmoScout.
> 
> Btw, I noticed a bug: When time plays backwards, the increase-time-speed-button does not what it's supposed to do. Is this bug introduced by these changes?

This is fixed for now, so both buttons behave the same. The question is, if the behaviour is accaptable as is or should the buttons just decrease speed in the opposite direction?

Another thing that I noticed is that the old hotkeys for doubling/halving time with `+`/`-` still work. This should be changed too.

Schneegans:
> This is fixed for now, so both buttons behave the same. The question is, if the behaviour is accaptable as is or should the buttons just decrease speed in the opposite direction?

This behaviour was the result of our user study. Most users expected that the buttons actually change the direction of time play back.

> Another thing that I noticed is that the old hotkeys for doubling/halving time with `+`/`-` still work. This should be changed too.

This should indeed be fixed. Maybe the hotkeys could act as you suggested?

Schneegans:
While testing the current state, I observered one issue: When you resize the window, most of the time the GUI gets not resized properly. This is especially apparent when toggling fullscreen mode with `Àlt-Return`.

This happens for me on Windows, I haven't tested it with Linux.

TheJhonny007:
> While testing the current state, I observered one issue: When you resize the window, most of the time the GUI gets not resized properly. This is especially apparent when toggling fullscreen mode with `Àlt-Return`.
> 
> This happens for me on Windows, I haven't tested it with Linux.

On Linux this problem doesn't exists. The GUI just needs a few milliseconds to get to the right size.

octfx:
> > While testing the current state, I observered one issue: When you resize the window, most of the time the GUI gets not resized properly. This is especially apparent when toggling fullscreen mode with `Àlt-Return`.
> > This happens for me on Windows, I haven't tested it with Linux.
> 
> On Linux this problem doesn't exists. The GUI just needs a few milliseconds to get to the right size.

I too can't confirm this. (Linux)

Schneegans:
And two additional feature requests 😄:
1. It would be nice if notification area moved down when the timeline is expanded.
2. If tabs in the sidebar are expanded such that the sidebar is very long, it gets cropped at the top and bottom which looks a bit buggy. I think its `div` should be expanded to the full available height and it should be z-positioned below the timeline and the statusbar. We could add a webkit mask gradient fading it to full transparency towards the edges. When the timeline is expanded, the mask should move down accordingly. Here's a quick mockup:

mockup

octfx:
> And two additional feature requests smile:
> 
>     ...
I'll look into both :+1:

octfx:
Notifications are now relative and move with the timeline height.

The sidebar now fades on the top and bottom. But I currently see no easy way to fully implement the mockup. :/

TheJhonny007:
> Notifications are now relative and move with the timeline height.
> 
> The sidebar now fades on the top and bottom. But I currently see no easy way to fully implement the mockup. :/

The status bar at the bottom now disappears when expanding the timeline.

octfx:
> > Notifications are now relative and move with the timeline height.
> > The sidebar now fades on the top and bottom. But I currently see no easy way to fully implement the mockup. :/
> 
> The status bar at the bottom now disappears when expanding the timeline.

See https://github.com/cosmoscout/cosmoscout-vr/pull/95/commits/79ae2ffd50cef3cd0004fa6728ca966fd20ee0af

Mask gradient has to be dynamically resized upon timeline expansion. WIP.

resources/gui/css/sidebar.css
```diff
@@ -1,27 +1,28 @@
-.side-bar {
+#sidebar {
     width: 450px;
-    position: absolute;
-    height: 100%;
+    height: 100vh;
+    margin-top: -25px;
     overflow-x: hidden;
```
Schneegans: Looks very nice! You could add `padding-bottom: 100px;` here in order to allow scrolling to the very end of the sidebar.

octfx: Padded things up!

Schneegans:
I "fixed" the gui-resize-issue in #103

TheJhonny007:
I implemented the linear interpolation for GuiAreas. They are always on for WorldSpaceGUIs and toggable for the main user interface with the default as off (looks much better on desktop and on a powerwall).

Schneegans:
I think we can solve the texture lookup / interpolation issue without the `pSmoothScreenSpaceGui` property. In e686892 I fixed the lookup position. The issue was that OpenGL samples at pixel centers - that means the top left pixel of a GuiArea will not have texture coordinates `[0, 0]` but rather `[0.5 / screenWidth, 0.5 / screenHeight]`.

With e686892 in place, nothing changes visually when `pSmoothScreenSpaceGui` is enabled or disabled for the main user interface. I believe it should be OK to just do no interpolation for ScreenSpaceGuiAreas and to always do linear Interpolation for WorldSpaceGuiAreas.

What do you think?

Schneegans:
I tested the current state of this branch and noticed a few regressions compared to `develop`:

- Changing the current month with the tiny arrow buttons or via the scroll wheel changes the time not by full months.
- The measurement tools buttons are missing their active state when selected.
- The image-channel dropdown of the lod-bodies-plugin is not correctly initialized when the active planet changes.

PR #99, opened 2019-12-25
Cleanup Shutdown

### LINKED ISSUE #98: Console window does not close on Windows
This happens from time to time.

Schneegans:
These changes make sure that all destructors are called properly at the end of the program execution. It should also fix #98.


TheJhonny007:
There are currently at least two circular dependencies with shared pointers which might prevent a complete cleanup:
```
cs::tools::Mark                <--> cs::core::GuiManager
cs::tools::MultiPointTool <--> cs::core::GuiManager
```

Schneegans:
A plugin which instantiates a Tool must store a reference to this Tool anyways (because it has to delete it one the plugin unloads). Once the Tool decides that it should be deleted (e.g. the user clicked its close button), we have to remove it from the plugin's list. Having the Tool stored in another list in the GuiManager or some kind of ToolManager is pretty pointless.

We could remove the comment - it's just a hint how a plugin which works with multiple Tools _could_ implement this - oftentimes Tools are managed in totally different ways. For example, the PathTool has several other Tools as members and keeps track of them itself. As it's a common source of errors in C++ when you try to delete items from a container while iterating over it, I thought that it might be useful...

TheJhonny007:
Yeah, I agree. I think it is fine for now.

PR #101, opened 2019-12-29
Comment Count Check for CI

Schneegans:
This PR adds the following features:
* For pull requests, the percentage of source code comments is compared to the base branch via Github Actions. If it decreased, the corresponding check is marked as failed. This way we can ensure that contributers try to add as many comments as possible.
* A new documentation page describing continuous integration for CosmoScout VR has been added.
* Version 2 of the checkout action is now used in our workflows.
* The layout of the footer navigation in the documentation pages has been slightly improved.


Schneegans:
Just a remark: The _Check Comment Percentage_ check of this PR fails because the `cloc.sh` of the base branch (`develop`) does not know the `--percentage-only` flag which gets introduced by this PR. So it cannot compare the amount of comments to the base branch 😃.

PR #112, opened 2020-02-03
Fix zoom issue

### LINKED ISSUE #74: Sometimes there is an instant-zoom-to-surface-issue
Especially at low frame rates, the observer moves much to quickly to a planet's surface.

Schneegans:
This fixes #74


Schneegans:
I will add some further comments to the code. Especially `SolarSystem::updateSceneScale()` and `SolarSystem::updateObserverFrame()` require some further explanations.

src/cosmoscout/ObserverNavigationNode.cpp
```diff
@@ -134,33 +134,42 @@ bool ObserverNavigationNode::DoEvalNode() {
...
-  vTranslation += vOffset * dObserverScale;
+  vTranslation += vOffset;
+
+  auto&  oObs     = mSolarSystem->getObserver();
+  double stepSize = glm::length(vTranslation);
+
+  if (stepSize > 0.0) {
+    // Ensure that an SolarSystem::updateSceneScale() is called at least at 100 Hz. If it is called
+    // only once a frame, it can happen that the observer instantly travels to a planet's surface.
+    int32_t steps = std::ceil(dDeltaTime / 0.01);
```
TheJhonny007: Why don't you simply multiply by 100? This would be a more in line with the 100 Hz in the comment above and it would also be more efficient (although that doesn't matter here).

Issue #109, opened 2020-01-28
Reduce the number of compiler warnings


Schneegans:
Especially MSVC produces a lot of warnings. In fact there are so many warnings that we have to ignore them currently.

- [x] MSVC `/W3 /WX`
- [x] GCC `-Wall -Werror`
- [x] clang `-Wall -Werror`
- [ ] clang-tidy

TheJhonny007:
Which warning level should we compile against? Microsoft recommends /W4, but default is /W3. 

I got the root repository to compile wit /W4 /WX, but there are some strange warnings in /W4 like unused parameters and stuff like that.

Schneegans:
In my opinion, a clean complete /W3 build would be quite awesome already. I think we should aim for that first and see much work /W4 would create afterwards.

TheJhonny007:
I think, if we want to keep `Werror` and `/WX` we need to integrate clang-tidy into the build chain. It catches most warnings independent of the compiler and prevents pushing code that doesn't work on other machines.

My initial attempt of integrating it failed, because it couldn't find the generated `*_export.hpp` header files.

Schneegans:
I pushed an initial `clang-tidy.sh` script: https://github.com/cosmoscout/cosmoscout-vr/blob/feature/clang-tidy/clang-tidy.sh

It fixes a lot, but also prints bazillions of warnings. And it takes quite some time. Do you have a good idea how this could be integrated in out CI? Running it on every commit like and doing a git-diff afterwards will be quite costly.

If we disable some checks, we could also strive for zero warnings...

TheJhonny007:
I will look into that monday. It is also possible to integrate clang-tidy with CMake easily.

CLion uses clang-tidy for it's suggestion and I'm quite happy with them. Maybe we can create a `.clang-tidy` file based on that. https://confluence.jetbrains.com/display/CLION/Clang-Tidy+in+CLion%3A+default+configuration

TheJhonny007:
I integrated clang-tidy with CMake. It can be enabled with `-DCS_RUN_CLANG_TIDY=On` (default is of). It will also be disabled if clang-tidy isn't installed.

I tried replicating CLions .clang-tidy config, but was unsuccessful.

I think we should leave it disabled on the CI for now and slowly over time reduce the warnings. We may start with a single suggestion library and work our way forward. But I would prefer if #119 would be done first.

TheJhonny007:
> And it takes quite some time.

[...]

PR #129, opened 2020-03-25
Removes all MSVC Warnings

TheJhonny007:
Fixes issue #109.

This pull requests deals with all level `/W3` compiler warnings that MSVC emits. I also fixed some `/W4` warnings in the core library.

From now on `/W3` warnings will be handled as errors.


src/cs-core/InputManager.hpp
```diff
@@ -35,6 +35,10 @@ class ScreenSpaceGuiArea;
 
 namespace cs::core {
 
+#ifdef _MSC_VER
+#pragma warning(disable : 4275)
+#endif _MSC_VER
```
TheJhonny007: Should we do something about this warning? It is currently the result of subclassing ViSTA classes that don't declare a DLL interface.

src/cs-graphics/internal/tiny_gltf_helper.hpp
```diff
@@ -92,9 +92,11 @@ auto find_material_parameter(
     return def;
   } else {
     auto const& parameter = it->second;
-    T           value;
+    T           value{};
     for (size_t i = 0; i < std::min(size_t(value.length()), parameter.number_array.size()); ++i) {
+#pragma warning(disable: 4267 4244)
       value[i] = parameter.number_array[i];
+#pragma warning(default: 4267 4244)
     }
```
Schneegans: I haven't tested it, but this should work as T is always a glm type:

```suggestion
    for (size_t i = 0; i < std::min(size_t(value.length()), parameter.number_array.size()); ++i) {
      value[i] = static_cast<typename T::value_type>(parameter.number_array[i]);
    }
```

src/cs-core/InputManager.hpp
Schneegans: If it just requires adding a missing `VISTAKERNELAPI` to the `IVistaDirectKeySink`, we should definitely do this.

src/cs-scene/CelestialAnchor.hpp
```diff
@@ -79,7 +79,7 @@ class CS_SCENE_EXPORT CelestialAnchor {
   virtual void   setAnchorScale(double dScale);
 
   /// Called regularly by the Universe if registered.
-  virtual void update(double tTime, CelestialObserver const& oObs) {
+  virtual void update(double, CelestialObserver const&) {
```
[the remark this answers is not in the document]
TheJhonny007: I don't like this solution. I split the declaration and assignment to avoid this warning. Also since this is a `/W4` warning we could in theory ignore it completely in the case we want to stay at `/W3`.

src/cs-core/InputManager.hpp
```diff
@@ -35,6 +35,13 @@ class ScreenSpaceGuiArea;
 
 namespace cs::core {
 
+// The warning can be ignored since the interface methods inhereted from ViSTA aren't used outside
+// this module.
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4275)
+#endif _MSC_VER
```
Schneegans:
```suggestion
#endif
```

src/cs-core/InputManager.hpp
```diff
@@ -169,4 +176,8 @@ class CS_CORE_EXPORT InputManager : public VistaKeyboardSystemControl::IVistaDir
 
 } // namespace cs::core
 
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif _MSC_VER
```
Schneegans:
```suggestion
#endif
```

Schneegans:
There are still a few unresolved things above. And will you add the fix to Vista?

TheJhonny007:
> There are still a few unresolved things above. And will you add the fix to Vista?

What fix for Vista?

Schneegans:
The `VISTAKERNELAPI` to the `IVistaDirectKeySink` in order to silence w4275.

And what do you think about the `static_cast<typename T::value_type>` in the tiny_gltf_helpers? In my opnion it's cleaner than to rely on the existence of a member `x`.

Schneegans:
I think there is a Linux build issue...

TheJhonny007:
> I think there is a Linux build issue...

Yeah it seems the Ubuntu images have some problems.
 Did you try it on a Linux machine already? Then I would proceed to merge anyways.

Schneegans:
`fopen_s` and `fscanf_s` are not supported!

TheJhonny007:
> `fopen_s` and `fscanf_s` are not supported!

But it's in the C11 standard: https://en.cppreference.com/w/c/io/fopen

Schneegans:
Yeah, but they seem to be only optional in the standard.

TheJhonny007:
> Yeah, but they seem to be only optional in the standard.

We can use `--std=C11` to enable it in GCC. I will try tomorrow as it is getting late and the CI servers aren't working anyways.

Schneegans:
Maybe we should `#define _CRT_SECURE_NO_DEPRECATE` in the affected files? See https://stackoverflow.com/questions/1513209/is-there-a-way-to-use-fopen-s-with-gcc-or-at-least-create-a-define-about-it

TheJhonny007:
> Maybe we should `#define _CRT_SECURE_NO_DEPRECATE` in the affected files? See https://stackoverflow.com/questions/1513209/is-there-a-way-to-use-fopen-s-with-gcc-or-at-least-create-a-define-about-it

The correct approach would probably to use an `ifstream`, but especially the scanf line looks horrible to implement with one.

Schneegans:
Yup. I think it's not worth the effort.

Schneegans:
It works with gcc now. Do you have an idea why the windows ci-build fails?

TheJhonny007:
> It works with gcc now. Do you have an idea why the windows ci-build fails?

No idea. It compiles on my machine. It seems that there is an issue with clcache.

Schneegans:
When inspecting the error message `PermissionError: [Errno 13] Permission denied: 'd:\\a\\cosmoscout-vr\\cosmoscout-vr\\install\\windows-externals-release/include'`, I suspect that it's related to `/external:I ${COSMOSCOUT_EXTERNALS_DIR}/include`. 

Maybe it should be `/external:I ${COSMOSCOUT_EXTERNALS_DIR}\\include`?

Schneegans:
BTW, have you tried `/external:anglebrackets`? This could work as well!

TheJhonny007:
> BTW, have you tried `/external:anglebrackets`? This could work as well!

We know in 90 minutes 😋

TheJhonny007:
By the way have you thought about caching the externals directly after building them, so a failed CosmoScout build after that doesn't required a rebuild of them?

TheJhonny007:
Now it seems like VS 2019 has more warnings than VS 2017... I will try using a newer version.

TheJhonny007:
I was unable to get the warning on my machine. I hope it was the only one the CI gets that I don't.

PR #131, opened 2020-03-27
Fix warnings

Schneegans:


src/cs-core/SolarSystem.cpp
```diff
@@ -499,17 +500,17 @@ void SolarSystem::printFrames() {
...
+  CS_DISABLE_GCC_WARNING("-Wwrite-strings")
 
   // Continue execution on errors.
   erract_c("SET", 0, "RETURN");
 
   // Disable default error reports.
   errdev_c("SET", 0, "NULL");
 
-#pragma GCC diagnostic pop
+  CS_WARNING_POP
```
TheJhonny007: We can avoid disabling warnings here by using:
```
erract_c("SET", 0, const_cast<char*>("RETURN"));
errdev_c("SET", 0, const_cast<char*>("NULL"));
```

src/cs-utils/utils.hpp
```diff
@@ -18,6 +18,20 @@
...
+#define CS_WARNING_PUSH _Pragma("GCC diagnostic push")
+#define CS_WARNING_POP _Pragma("GCC diagnostic pop")
+#define CS_DISABLE_GCC_WARNING(warningName) CS_DO_PRAGMA(GCC diagnostic ignored warningName)
+#define CS_DISABLE_MSVC_WARNING(warningNumber)
+#elif defined(_MSC_VER)
+#define CS_WARNING_PUSH __pragma(warning(push))
+#define CS_WARNING_POP __pragma(warning(pop))
+#define CS_DISABLE_MSVC_WARNING(warningNumber) __pragma(warning(disable : warningNumber))
+#define CS_DISABLE_GCC_WARNING(warningName)
+#endif
```
TheJhonny007: I would use plural here: `WARNINGS`

PR #137, opened 2020-04-06
Feature/simple wms bodies

gmate321:


config/base/scene/simple_desktop.json
```diff
@@ -241,6 +235,79 @@
...
         }
       }
+    },
+	"csp-simple-wms-bodies": {
+      "bodies": {
+        "Earth": {
+          "texture":  "../share/resources/textures/earth.jpg",
+          "wms": [
+            {
+              "name": "Fires",
```
Flatken, line 245: Maybe extend the name with sensor sources (MODIS?, VIIRES? sensor) .e.g. Fire-Hotspots-MODIS

config/base/scene/simple_desktop.json
```diff
@@ -241,6 +235,79 @@
...
+              "copyright": "NASA",
+              "url": "https://neo.sci.gsfc.nasa.gov/wms/wms?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&CRS=CRS:84&BBOX=-180,-90,180,90&FORMAT=image/png",
+              "width": 1600,
+              "height": 800,
+              "time": "2000-02-01/2019-11-01/P1M",
+              "layers": "MOD_LSTD_M",
+              "preFetch": 5
+            },
+            {
+              "name": "angeeichtes Radarkomposit (RW)",
```
Flatken, line 275: maybe an english name is better

config/base/scene/simple_desktop.json
```diff
@@ -241,6 +235,79 @@
...
+              "copyright": "NASA",
```
Flatken, line 276: This is DWD? Deutscher Wetterdienst?

config/base/scene/simple_desktop.json
```diff
@@ -241,6 +235,79 @@
...
+              "url": "https://maps.dwd.de/geoserver/ows?service=wms&version=1.3.0&request=GetMap&CRS=EPSG:1000004&BBOX=-180,-90,180,90&FORMAT=image/png",
+              "width": 1600,
+              "height": 800,
+              "time": "2020-01-07T08:30:00.000Z/2020-01-08T08:30:00.000Z/PT10M",
+              "layers": "dwd:RADOLAN-RW",
+              "preFetch": 5
+            },
+            {
+              "name": "Radarvorhersage (Basis RX)",
```
Flatken, line 285: Same here: Give an english name

config/base/scene/simple_desktop.json
```diff
@@ -241,6 +235,79 @@
...
+              "copyright": "DWD",
+              "url": "https://maps.dwd.de/geoserver/ows?service=wms&version=1.3.0&request=GetMap&CRS=EPSG:1000004&BBOX=-180,-90,180,90&FORMAT=image/png",
+              "width": 1600,
+              "height": 800,
+              "time": "2020-01-08T08:45:00.000Z,2020-01-08T08:50:00.000Z,2020-01-08T08:55:00.000Z,2020-01-08T09:00:00.000Z,2020-01-08T09:05:00.000Z,2020-01-08T09:10:00.000Z,2020-01-08T09:15:00.000Z,2020-01-08T09:20:00.000Z,2020-01-08T09:25:00.000Z,2020-01-08T09:30:00.000Z,2020-01-08T09:35:00.000Z,2020-01-08T09:40:00.000Z,2020-01-08T09:45:00.000Z,2020-01-08T09:50:00.000Z,2020-01-08T09:55:00.000Z,2020-01-08T10:00:00.000Z,2020-01-08T10:05:00.000Z,2020-01-08T10:10:00.000Z,2020-01-08T10:15:00.000Z,2020-01-08T10:20:00.000Z,2020-01-08T10:25:00.000Z,2020-01-08T10:30:00.000Z,2020-01-08T10:35:00.000Z,2020-01-08T10:40:00.000Z,2020-01-08T10:45:00.000Z,2020-01-08T10:50:00.000Z",
+              "layers": "dwd:FX-Produkt"
+            }
+          ]
+        },
+        "Mars": {
```
Flatken, line 295: Delete the Mars section here and add to simple-bodies. The reason is that this server is DLR internal and external people do not have access to that one

resources/gui/js/apis/timeline.js
```diff
@@ -368,6 +368,11 @@ class TimelineApi extends IApi {
     this._itemsOverview.update(data);
   }
 
+  removeItem(id) {
```
Flatken, line 371: What happens if id does not exist?
Do we need to catch that?

Flatken, line 371: Do we need to check if the id does exist?

config/base/scene/simple_desktop.json
gmate321: Where do I find this sensor sources?

config/base/scene/simple_desktop.json
gmate321: Yes, thank you for noticing!

resources/gui/js/apis/timeline.js
gmate321: No, "The method ignores removal of non-existing items, and returns an array containing the ids of the items which are actually removed from the DataSet." (https://ww3.arb.ca.gov/ei/tools/lib/vis/docs/dataset.html#Remove)

Schneegans:
Thanks for the work of porting this to the newest CosmoScout API! I added some remarks in https://github.com/cosmoscout/csp-simple-wms-bodies/pull/3. In addition, there are several clang / gcc warnings which are treated as errors.

gmate321:
> Great work!
> 
> Can you also expand the plugin documentation in the README?

Thank you! What would you like to see in the README?

TheJhonny007:
> > Great work!
> > Can you also expand the plugin documentation in the README?
> 
> Thank you! What would you like to see in the README?

A link and a short description with an image like all the other default plugins here:
https://github.com/cosmoscout/cosmoscout-vr#plugins-for-cosmoscout-vr

Schneegans:
Can anybody else test this on Linux? It still fails to download any images on my side...

config/base/scene/simple_desktop.json
```diff
@@ -241,6 +238,67 @@
           "texture": "../share/resources/textures/neptune.jpg"
         }
       }
+    },
+	"csp-simple-wms-bodies": {
```
Flatken, line 242: formatting looks strange? Tabs!? Meybe only here on github?

Flatken, line 242: formatting looks strange? Tabs!? Meybe only here on github?

config/base/scene/simple_desktop.json
Flatken: https://firms.modaps.eosdis.nasa.gov/

Flatken:
Please look for the conflicts!

config/base/scene/simple_desktop.json
TheJhonny007: It is indentet with a tab instead of spaces.

TheJhonny007:
I let GitHub auto merge develop into your branch. Some code broke since then, because develop changed a lot since this pull request was issued.

gmate321:
> 
> 
> Can anybody else test this on Linux? It still fails to download any images on my side...

It works for me now on Linux.

TheJhonny007:
Looks great! I think we can merge it soon.

Here are some thing I noticed:
- The _Fires-Hotspots-VIIRS_ dataset crashes, when I fast forward/backward a few months.
- The errors should be more descriptive. If a dataset doesn't have data for the current time print that information as a warning instead of `Failed to load ...`.
- Make it possible to select `none` as a source to disable the plugin temporarily.

gmate321:
> 
> 
> Looks great! I think we can merge it soon.
> 
> Here are some thing I noticed:
> 
>     * The _Fires-Hotspots-VIIRS_ dataset crashes, when I fast forward/backward a few months.
> 
>     * The errors should be more descriptive. If a dataset doesn't have data for the current time print that information as a warning instead of `Failed to load ...`.
> 
>     * Make it possible to select `none` as a source to disable the plugin temporarily.

What did you notice with the Fires-Hotspots-VIIRS dataset, how did it crash?

TheJhonny007:
> What did you notice with the Fires-Hotspots-VIIRS dataset, how did it crash?

If I scroll through time in the timeline the application just closes.
The crash report is this line: `Process finished with exit code -1073741819 (0xC0000005)`

TheJhonny007:
So the line that crashes is the following: [`SimpleWMSBody.cpp:386`](https://github.com/cosmoscout/cosmoscout-vr/blob/feature/simple-wms-bodies/plugins/csp-simple-wms-bodies/src/SimpleWMSBody.cpp#L386)

gmate321:
> 
> 
> So the line that crashes is the following: [`SimpleWMSBody.cpp:386`](https://github.com/cosmoscout/cosmoscout-vr/blob/feature/simple-wms-bodies/plugins/csp-simple-wms-bodies/src/SimpleWMSBody.cpp#L386)

Thanks, I will investigate that.

Schneegans:
On which Linux did you test this? For me it fails to download the images due to SSL problems. If I put

```c++
request.setOpt(curlpp::options::SslVerifyPeer(false));
```
into `WebMapTextureLoader.cpp`, the fire dataset is downloaded properly, for the NASA server I still get `error:1414D172:SSL routines:tls12_check_peer_sigalg:wrong signature type`.

gmate321:
> 
> 
> On which Linux did you test this? For me it fails to download the images due to SSL problems. If I put
> 
> ```c++
> request.setOpt(curlpp::options::SslVerifyPeer(false));
> ```
> 
> into `WebMapTextureLoader.cpp`, the fire dataset is downloaded properly, for the NASA server I still get `error:1414D172:SSL routines:tls12_check_peer_sigalg:wrong signature type`.

On Ubuntu 18.04.4. This SSL error never happened to me.

Schneegans:
Could you execute this on your Linux machine and post the results?

```bash
cd install/linux-Release/bin
export LD_LIBRARY_PATH=../lib:../lib/DriverPlugins:$LD_LIBRARY_PATH
ldd cosmoscout | grep curl
ldd cosmoscout | grep ssl
```

gmate321:
> ```shell
> export LD_LIBRARY_PATH=../lib:../lib/DriverPlugins:$LD_LIBRARY_PATH
> ```
Here are the results:
```
ldd cosmoscout | grep curl
	libcurl.so => ../lib/libcurl.so (0x00007f842968f000)
	libcurlpp.so.1 => ../lib/libcurlpp.so.1 (0x00007f8429478000)
ldd cosmoscout | grep ssl
	libssl.so.1.1 => /lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007f5417b19000)
```

Schneegans:
I just realized that we use `request.setOpt(curlpp::options::SslVerifyPeer(false));` also for downloading files at application startup. So maybe we should fix our curl/ssl-setup in general. I think that we need something like `request.setOpt(curlpp::options::CaPath("/etc/ssl/certs"));` to make libcurl use the system's certificates. I am not sure how it finds these certificates on your end though... I tested it on several machines now and on neither of them the images got downloaded.

But this seems out of the scope of this PR, so I would suggest to add `request.setOpt(curlpp::options::SslVerifyPeer(false));` for this request and then we should open a new issue for the proper SSL handling.

Here are some other issues I came across:
* Unloading the plugin does not clean the UI. Try executing `CosmoScout.callbacks.core.reloadPlugin("csp-simple-wms-bodies")` - you'll end up with twice as many side-bar items.
* I would remove the feature of showing the time ranges in the timeline. In it's current state it's pretty confusing & buggy. What are the semantics of editing such a range? When you save the current scene (`CosmoScout.callbacks.core.save("test.json")`), you will actually save all those temporary bookmarks as well. And later, when you load the scene, your timeline will become totally cluttered. We would need some kind of timeline API to add some subtle region highlights for this to be implemented properly. Until then, I would suggest to drop this feature.
* I think it would great to have some default true-color imagery dataset. Like this: https://neo.sci.gsfc.nasa.gov/wms/wms?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&CRS=CRS:84&BBOX=-180,-90,180,90&FORMAT=image/png&WIDTH=1600&HEIGHT=800&LAYERS=VIIRS_543D. This however, crashes for me...? The config I used is below.
* The fire dataset has some API key in it. Are we allowed to publish this?
* It would be great to support jpeg as a image format, as this would significantly speed up the file downloads!

```json
"VIIRS_543D": {
      "copyright": "NASA",
      "url": "https://neo.sci.gsfc.nasa.gov/wms/wms?SERVICE=WMS&REQUEST=GetMap&VERSION=1.3.0&CRS=CRS:84&BBOX=-180,-90,180,90&FORMAT=image/png&WIDTH=1600&HEIGHT=800",
      "width": 1600,
      "height": 800,
      "time": "2014-01-17/2014-02-25/P1D,2014-02-27/2014-03-04/P1D,2014-03-06/2019-11-20/P1D,2019-11-22/2020-06-08/P1D,2020-06-11/2020-06-20/P1D",
      "layers": "VIIRS_543D",
      "preFetch": 2
},
```

gmate321:
Thanks for finding these issues. The plugin unload is no fixed and I've removed the bookmarks for now. Regarding the other issues:

- After some testing I could also reproduce the crash due to SSL problems, but it happened rarely to me. I've mostly found it when I fast forwarded or rewinded a lot, which produced a lot of WMS requests. After some time some of these requests failed.  It only happenedet with the NASA NEO datasets (true color dataset and the two land surface temperatures) though.
- I am not sure about the key in the fires dataset. I think there is a 300 requests / 10 minutes limit per key (https://firms.modaps.eosdis.nasa.gov/web-services/).
- Do you imagine jpeg support with a definition per dataset in startup stettings?
- The interpolation was indeed not working for the past two weeks. The duration of timesteps were previously simply defined in seconds, which is fine for datasets with refresh rates given in days, but not for monthly / yearly dataset. I've temporarily let the time string formatting deal with finding the previous timestep. Incrementing and interpolation was not possible this way. Now I reworked the timestep duration in `utils.hpp` to store years and months separately and it works with all refresh rates. Time interpolation is also back. It is assumed that a monthly dataset sample is always taken on the 1st day of the month. Should we expect datasets which don't follow this?

Schneegans:
This will soon be merged in form of https://github.com/cosmoscout/cosmoscout-vr/pull/237.

PR #138, opened 2020-04-07
Feature/clang tidy

### LINKED ISSUE #109: Reduce the number of compiler warnings

JonasGilg:
This pull requests adds clang-tidy checks to the project and resolves all warnings emited by clang-tidy.

Currently the following check suites are enabled:
- bugprone
- cert
- clang-analyzer
- cppcoreguidelines
- google
- hicpp
- misc
- modernize
- performance
- portability
- readability

Some warnings are excluded:
- `clang-diagnostic-inconsistent-dllimport`: I don't know enough about DLLs yet to fix this warning
- `bugprone-exception-escape`: `spdlog` can theoretically throw exceptions which is used in a lot of places.
- `bugprone-narrowing-conversions`: Duplicate warning.
- `cert-oop54-cpp`: Duplicate warning.
- `cppcoreguidelines-avoid-c-arrays`: Duplicate warning.
- `cppcoreguidelines-avoid-magic-numbers`: We have way to many magic numbers for this to be enabled currently.
- `cppcoreguidelines-explicit-virtual-functions`: Dublicate warning.
- `cppcoreguidelines-macro-usage`: Because of the DLL-Exports.
- `cppcoreguidelines-non-private-member-variables-in-classes`: I don't see the problem with public members.
- `cppcoreguidelines-pro-type-vararg`: Library interoperability.
- `cppcoreguidelines-pro-type-union-access`: GLM requires this to be disabled.
- `google-explicit-constructor`: Dupicate warning.
- `google-readability-todo`: Don't see, why TODOs need a user attached. There is `git blame` for that.
- `google-readability-braces-around-statements`: Dupicate warning.
- `google-runtime-references`: `nlohmann::json` needs them a lot.
- `hicpp-avoid-c-arrays`: Duplicate warning.
- `hicpp-braces-around-statements`: Duplicate warning.
- `hicpp-member-init`: Duplicate warning.
- `hicpp-noexcept-move`: Duplicate warning.
- `hicpp-no-array-decay`: Duplicate warning.
- `hicpp-signed-bitwise`: Duplicate warning.
- `hicpp-special-member-functions`: Duplicate warning.
- `hicpp-uppercase-literal-suffix`: Duplicate warning.
- `hicpp-use-equals-default`: Duplicate warning.
- `hicpp-use-override`: Duplicate warning.
- `hicpp-vararg`: Library interoperability (Duplicate warning).
- `misc-non-private-member-variables-in-classes`: See above.
- `modernize-use-nodiscard`: Clutters code too much.
- `modernize-use-trailing-return-type`: Personal preference.
- `readability-implicit-bool-conversion`: Personal preference.
- `readability-magic-numbers`: We have way to many magic numbers for this to be enabled currently.

There are also some warnings enabled, but they are not treated as errors:
- `cert-env33-c`: These can be unavoidable some times, but shouldn't be the used liberally.
- `cert-err58-cpp`: These should be avoided, but sometimes they are required.
- `clang-analyzer-cplusplus.Move`: I would like to treat them as errors, but there are false positives with `spdlog`.
- `clang-analyzer-cplusplus.NewDeleteLeaks`: I would like to treat them as errors, but there are false positives with `spdlog`.
- `cppcoreguidelines-pro-type-reinterpret-cast`: Is unavoidable sometimes.
- `google-default-arguments`: Needs discussion.

The above checks are open for discussion. I also started fixing a lot of them at the start before excluding them, so some parts of the code might already be conforming to disabled checks.


gabime:
>  spdlog can theoretically throw exceptions 

No it is not. spdlog only throws when constructing new loggers/sinks, not while using

TheJhonny007:
> > spdlog can theoretically throw exceptions
> 
> No it is not. spdlog only throws when constructing new loggers/sinks, not while using

When having an `spdlog::debug` or similar in a destructor, clang-tidy is giving me this error: `bugprone-exception-escape`. I don't know why (that is why I wrote theoretically), but it is there.

Example: https://github.com/cosmoscout/cosmoscout-vr/blob/feature/clang-tidy/src/cs-core/SolarSystem.cpp#L58

gabime:
Which version of spdlog are you using ?

TheJhonny007:
> Which version of spdlog are you using ?

1.5

.clang-tidy
```diff
@@ -0,0 +1,64 @@
...
+
+WarningsAsErrors: '*,
+-cert-env33-c,
+-cert-err58-cpp,
+-clang-analyzer-cplusplus.Move,
+-clang-analyzer-cplusplus.NewDeleteLeaks,
+-cppcoreguidelines-pro-type-reinterpret-cast,
+-google-default-arguments'
+
+HeaderFilterRegex: '.*/csp?-.*\.hpp'
```
Schneegans, line 53: What is the purpose of this filter? Do we really need it?

CMakeLists.txt
```diff
@@ -248,6 +254,46 @@ else()
...
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${CMAKE_BINARY_DIR}/src/cs-utils")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${CMAKE_BINARY_DIR}/src/cs-scene")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${CMAKE_BINARY_DIR}/src/cs-graphics")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${CMAKE_BINARY_DIR}/src/cs-core")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${CMAKE_BINARY_DIR}/src/cs-gui")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${COSMOSCOUT_EXTERNALS_DIR}/include")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${COSMOSCOUT_EXTERNALS_DIR}/include/cef")
+    set(CS_CT_OPT "${CS_CT_OPT};-extra-arg=-I${Boost_INCLUDE_DIRS}")
+
+    set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};${CS_CT_OPT}")
```
Schneegans, line 285: At least on Linux this is sufficient, as we create the `compile_commands.json` in `${CMAKE_BINARY_DIR}`. I have to test this on Windows though.

```suggestion
    set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};-quiet;-p;${CMAKE_BINARY_DIR}")
```

src/cosmoscout/Application.cpp
```diff
@@ -738,30 +747,34 @@ void Application::closePlugin(std::string const& name) {
...
-        auto polar = cs::utils::convert::toLngLatHeight(intersection.mPosition, radii[0], radii[0]);
-        auto lngLat = cs::utils::convert::toDegrees(polar.xy());
-
-        if (!std::isnan(lngLat.x) && !std::isnan(lngLat.y) && !std::isnan(polar.z)) {
-          mGuiManager->getGui()->executeJavascript(
-              fmt::format("CosmoScout.state.pointerPosition = [{}, {}, {}];", lngLat.x, lngLat.y,
-                  polar.z / mGraphicsEngine->pHeightScale.get()));
-          return;
+  mInputManager->pHoveredObject.connect(
+      [this](const cs::core::InputManager::Intersection& intersection) {
```
Schneegans, line 751: We usually use `Type const& name` in our code base. I know that `const Type& name` is more commonly used, but for the sake of consistency we should stick to `Type const& name`. Changing this would mean changing every source file. If we want to do this, this should be done in a separate PR.

src/cs-core/GuiManager.cpp
```diff
@@ -397,9 +398,10 @@ void GuiManager::addCssToGui(const std::string& fileName) {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-void GuiManager::addEventToTimenavigationBar(std::string start, std::optional<std::string> end,
-    std::string id, std::string content, std::optional<std::string> style, std::string description,
-    std::string planet, std::string place) {
+void GuiManager::addEventToTimenavigationBar(const std::string& start,
+    const std::optional<std::string>& end, const std::string& id, const std::string& content,
+    const std::optional<std::string>& style, const std::string& description,
+    const std::string& planet, const std::string& place) {
```
Schneegans, line 404:
```suggestion
void GuiManager::addEventToTimenavigationBar(std::string const& start,
    std::optional<std::string> const& end, std::string const& id, std::string const& content,
    std::optional<std::string> const& style, std::string const& description,
    std::string const& planet, std::string const& place) {
```

src/cs-core/GuiManager.hpp
```diff
@@ -141,9 +148,9 @@ class CS_CORE_EXPORT GuiManager {
   /// @param description The description of the event.
   /// @param planet Planet the event is happening on.
   /// @parama place The location on the planet.
-  void addEventToTimenavigationBar(std::string start, std::optional<std::string> end,
-      std::string id, std::string content, std::optional<std::string> style,
-      std::string description, std::string planet, std::string place);
+  void addEventToTimenavigationBar(const std::string& start, const std::optional<std::string>& end,
+      const std::string& id, const std::string& content, const std::optional<std::string>& style,
+      const std::string& description, const std::string& planet, const std::string& place);
```
Schneegans, line 153:
```suggestion
  void addEventToTimenavigationBar(std::string const& start, std::optional<std::string> const& end,
      std::string const& id, std::string const& content, std::optional<std::string> const& style,
      std::string const& description, std::string const& planet, std::string const& place);
```

src/cs-utils/Signal.hpp
```diff
@@ -53,14 +62,14 @@ class Signal {
 
   /// Calls all connected functions.
   void emit(Args... p) {
-    for (auto it : mSlots) {
+    for (const auto& it : mSlots) {
```
Schneegans, line 65:
```suggestion
    for (auto const& it : mSlots) {
```

src/cs-utils/Signal.hpp
```diff
@@ -53,14 +62,14 @@ class Signal {
...
       it.second(p...);
     }
   }
 
   /// Calls all connected functions except for one.
   void emitForAllButOne(int excludedConnectionID, Args... p) {
-    for (auto it : mSlots) {
+    for (const auto& it : mSlots) {
```
Schneegans, line 72:
```suggestion
    for (auto const& it : mSlots) {
```

CMakeLists.txt
```diff
@@ -248,6 +254,46 @@ else()
   )
 endif()
 
+# clang-tidy ---------------------------------------------------------------------------------------
+
+option(CS_RUN_CLANG_TIDY "Run clang-tidy with the compiler." Off)
```
Schneegans, line 259: For the sake of consistency, I would suggest to call this option `COSMOSCOUT_RUN_CLANG_TIDY`.

Schneegans (changes requested):
That's quite a lot to go through... but apart from the few things I remarked / changed it looks very good.

.clang-tidy
TheJhonny007: It makes sure that all header files in a path with "cs-" or "csp-" are checked. I found that to be the most reliable way of having headers included in the checks.

Do you have a better method?

src/cosmoscout/Application.cpp
TheJhonny007: Yes, we should stick to east-const. This looks like the result of `clang-tidy` with the `-fix` option. Maybe we should disable it completely and leave the fixing to the developer?

CMakeLists.txt
Schneegans: Did you get this working with the Visual Studio CMake generator? Or are you using Ninja?

TheJhonny007: On Windows it ran with MSBuild ("Visual Studio 16 2019" -A x64) and NMake ("CodeBlocks - NMake Makefiles"). Ninja is a little tricky with the MSVC toolchain.

Schneegans:
Did you get the `CS_RUN_CLANG_TIDY` option to work with the default Visual Studio CMake Generator? For me it doesn't work somehow. I think it should work with Ninja though - and then you could also supply the `-DCMAKE_EXPORT_COMPILE_COMMANDS=On` and use only `set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};-quiet;-p;${CMAKE_BINARY_DIR}")` as I suggested above. Can you confirm this?

TheJhonny007:
> Did you get the `CS_RUN_CLANG_TIDY` option to work with the default Visual Studio CMake Generator? For me it doesn't work somehow. I think it should work with Ninja though - and then you could also supply the `-DCMAKE_EXPORT_COMPILE_COMMANDS=On` and use only `set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};-quiet;-p;${CMAKE_BINARY_DIR}")` as I suggested above. Can you confirm this?

You are right, MSBuild (default Visual Studio CMake Generator) can't run clang-tidy. 

With NMake I get the following error if I remove the extra-args, even though `CMAKE_EXPORT_COMPILE_COMMANDS` is enabled:

```
C:\C++\cosmoscout-vr\src\cs-utils/CommandLine.hpp:10:10: error: 'cs_utils_export.hpp' file not found [clang-diagnostic-error]
#include "cs_utils_export.hpp"
...
```

I'll have a look at Ninja, but currently it wants to use MinGWs gcc as a compiler...

Schneegans:
Did the NMake generator create the file `build\windows-release\compile_commands.json`?

TheJhonny007:
> Did the NMake generator create the file `build\windows-release\compile_commands.json`?

Yes.

Schneegans:
Could you try running clang-tidy manually? It works like this on Linux:

```bash
clang-tidy -quiet -p build\windows-release src\cs-utils\CommandLine.hpp
```

TheJhonny007:
> Could you try running clang-tidy manually? It works like this on Linux:
> 
> ```shell
> clang-tidy -quiet -p build\windows-release src\cs-utils\CommandLine.hpp
> ```

That strangely works. 

I enabled verbose makefiles for debugging and it looks as if cmake invokes the same command:

```
cd C:\C++\cosmoscout-vr\build\windows-release\src\cs-utils

<path_to_cmake>\cmake.exe -E __run_co_compile --tidy="<path_to_clang-tidy>/clang-tidy.exe;-quiet;-p;C:/C++/cosmoscout-vr/build/windows-release" --source=C:\C++\cosmoscout-vr\src\cs-utils\CommandLine.cpp -- C:\PROGRA~2\MIB055~1\2019\COMMUN~1\VC\Tools\MSVC\1425~1.286\bin\Hostx64\x64\cl.exe @C:\Users\<user_name>\AppData\Local\Temp\nm6827.tmp

C:\C++\cosmoscout-vr\src\cs-utils/CommandLine.hpp:10:10: error: 'cs_utils_export.hpp' file not found [clang-diagnostic-error]
#include "cs_utils_export.hpp"
```

Schneegans:
Except for the forward slashes...?

TheJhonny007:
> Except for the forward slashes...?

Looks like clang-tidy can deal with both. If I run:

```
clang-tidy -quiet -p C:/C++/cosmoscout-vr/build/windows-release C:/C++/cosmoscout-vr/src/cs-utils/CommandLine.cpp
```

it works too. I can even mix them within the same path.

Schneegans:
Maybe we drop the idea of integrating clang-tidy with CMake?
* On Windows only Ninja and NMake Generators seem to be supported
* On Linux it's very slow - I would only enable it occasionally and in those cases I can use the clang-tidy.sh script which we could also integrate in our CI
* Most modern IDEs support the .clang-tidy file anyways and will display proper linting

What do you think?

TheJhonny007:
> Maybe we drop the idea of integrating clang-tidy with CMake?
> 
> * On Windows only Ninja and NMake Generators seem to be supported
> * On Linux it's very slow - I would only enable it occasionally and in those cases I can use the clang-tidy.sh script which we could also integrate in our CI
> * Most modern IDEs support the .clang-tidy file anyways and will display proper linting
> 
> What do you think?

I got it working with Ninja. NMake seems to be bugged and needs the `extra-args`.

The advantage to have it run side by side with `cl.exe` is that it catched some errors that were dependend on the msvc compiler options. But In general I agree with you. It currently only works with Ninja and getting that to run using MSVC with MinGW on the same machine is work enough.

Schneegans:
I would like to merge it rather soonishly as it will create a lot of merge conflicts with other branches. Do you create PRs for all the submodules? Or do we want to fix their errors later?

TheJhonny007:
> I would like to merge it rather soonishly as it will create a lot of merge conflicts with other branches. Do you create PRs for all the submodules? Or do we want to fix their errors later?

I can do that now.

Schneegans (approved):

PR #139, opened 2020-04-09
Fast build

TheJhonny007:
Requires: 
- https://github.com/cosmoscout/opensg-1.8/pull/1
- https://github.com/cosmoscout/vista/pull/3

### Improves OpenSG and ViSTA compile times:
If the current CMAKE_VERSION is at least 3.16 unity builds and precompiled headers will be used. This reduced the OpenSG compile time on Windows from 15 Minutes to 2 Minutes on my machine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,7 +91,6 @@ jobs:
           sudo apt-get update -q
           sudo apt-get install libc++-dev libgl1-mesa-dev libglu1-mesa-dev libx11-dev
           sudo apt-get install libxi-dev libgconf-2-4 libboost1.70-dev ccache lcov xvfb
-          sudo pip install cmake==3.13.3
           ccache --max-size 1G
           ccache -z
       # Workaround for https://github.com/actions/virtual-environments/issues/10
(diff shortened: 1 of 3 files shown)


Schneegans:
Did you try enabling unity builds and precompiled headers for CosmoScout itself?

TheJhonny007:
> Did you try enabling unity builds and precompiled headers for CosmoScout itself?

Not yet. I wanted to get the dependencies out of the way first, since they take the longest.

Schneegans:
Alright. Looks great so far! Only the one issue I pointed out in the Vista-PR. From my side everything seems good, it also seems to work nicely with ccache.

TheJhonny007:
Alright, it should work now. I don't know what is wrong with the clang build. Maybe it is a caching error?

Schneegans:
Maybe this is related: https://gitlab.kitware.com/cmake/cmake/issues/19721? We could disable the precompiled headers for the clang build for now...

TheJhonny007:
> Maybe this is related: https://gitlab.kitware.com/cmake/cmake/issues/19721? We could disable the precompiled headers for the clang build for now...

They also suggest, that it might be linked to cache. I disabled them for now when clang is used.

TheJhonny007:
I think it is ready now. Maybe you can run CosmoScout on Linux once with this branch @Schneegans?

PR #143, opened 2020-04-21
[Vestec] Fix building externals on windows

octfx:
Patched building curlpp using `make_externals.bat` from develop.  
Removed duplicate gdal building step


octfx:
Is fixed in an unpushed commit

PR #148, opened 2020-04-27
Update GUI

Schneegans:
This PR makes sure that all interface widgets are update when the corresponding C++-Properties are changed. This is especially useful when loading the scene-state from file.

### DIFF
--- a/src/cosmoscout/Application.cpp
+++ b/src/cosmoscout/Application.cpp
@@ -1119,7 +1119,7 @@ void Application::registerGuiCallbacks() {
         mSettings->mGraphics.pAmbientBrightness = static_cast<float>(std::pow(val, 10.0));
       }));
   mSettings->mGraphics.pAmbientBrightness.connect([this](float val) {
-    mGuiManager->setSliderValue("graphics.setAmbientLight", std::pow(val, 0.1));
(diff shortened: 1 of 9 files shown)


src/cosmoscout/Application.cpp
```diff
@@ -1086,12 +1118,9 @@ void Application::registerGuiCallbacks() {
...
         mSettings->mGraphics.pAmbientBrightness = static_cast<float>(std::pow(val, 10.0));
       }));
-
-  // Adjusts the amount of artificial glare in HDR mode.
-  mGuiManager->getGui()->registerCallback("graphics.setGlowIntensity",
-      "Adjusts the amount of glow of overexposed areas.", std::function([this](double val) {
-        mSettings->mGraphics.pGlowIntensity = static_cast<float>(val);
-      }));
+  mSettings->mGraphics.pAmbientBrightness.connect([this](float val) {
+    mGuiManager->setSliderValue("graphics.setAmbientLight", std::pow(val, 0.1));
```
TheJhonny007, line 1122:
```suggestion
    mGuiManager->setSliderValue("graphics.setAmbientLight", std::pow(val, 0.1F));
```

PR #149, opened 2020-04-28
HTTP API for CosmoScout

Schneegans:
I will add some documentation, but you can already start to play around! Simply checkout the branch, compile the externals and CosmoScout, launch it, and navigate to http://localhost:9001.

See https://github.com/cosmoscout/csp-web-api/pull/1.

Edit: The plugin is not loaded per default now. You have to add this to you settings:

```json
"csp-web-api": {
  "port": 9001,
  "page": "../share/resources/gui/example-web-frontend.html"
},
```


TheJhonny007:
On Windows I was promptet for admin privileges after (or at the end) of plugin loading. I denied the request and the applications seems to work normally.

I can't reproduce it, since Windows seems to remember the decision. (I will try and work around that).

make_externals.bat
```diff
@@ -174,6 +174,18 @@ cmake %CMAKE_FLAGS% -DCMAKE_INSTALL_PREFIX="%INSTALL_DIR%"^
...
 
+rem civetweb -----------------------------------------------------------------------------------------
+
+echo.
+echo Building and installing civetweb ...
+echo.
+
+cmake -E make_directory "%BUILD_DIR%/civetweb" && cd "%BUILD_DIR%/civetweb"
+cmake %CMAKE_FLAGS% -DCMAKE_INSTALL_PREFIX="%INSTALL_DIR%" -DCIVETWEB_ENABLE_CXX=On ^
+      -DBUILD_SHARED_LIBS=On -DSPDLOG_ENABLE_PCH=On "%EXTERNALS_DIR%/civetweb" || exit /b
```
TheJhonny007, line 185: Is this right?

TheJhonny007:
This plugin introduces a new external dependency that has to be installed for the whole of CosmoScout for a feature that is really specialized and only used within this plugin.

Do you think we can add dependencies of plugins that are only required by that plugin differently?
The first thing coming to my mind is using CMake's exernal projects feature: https://cmake.org/cmake/help/latest/module/ExternalProject.html

Schneegans:
This is a difficult question. With dependencies such as `civetweb` which can be build easily and do not drag in other things, I am quite happy to treat it like it's done now. This is especially true once we put the core plugins into the main repository. And it can happen that we use the library in other plugins in the future - `curl`, `curlpp` and `cares` were only required by the csp-lod-bodies plugin when we introduced it and now they are required by multiple plugins and `cs-utils`.

However there are more complicated dependencies (like VTK in the scivis branch). The best solution would be along these lines:
* Get rid of `make_externals.sh` and `make_externals.bat`
* In the main CMakeLists.txt we would call `ExternalProject_Add` for each core dependency.
* Each plugin would call `ExternalProject_Add` for *all* additional dependencies. This however can lead to multiple calls to `ExternalProject_Add` for the same dependency. This would need to be resolved, especially if multiple plugins request different versions of the same dependency.
* It also has to work for external dependencies which do not use CMake

What do you think? Would this be possible? But it's definitely not in the scope of this pull request...

TheJhonny007:
Moved this discussion to an issue: #150

Schneegans:
I think I will remove the plugin from the default config, as it's not required for standard use cases. Are there any other ideas?

PR #151, opened 2020-04-28
Feature/fix dropdown

gmate321:
Event listeners were added to every `simple-value-dropdown` multiple times, since initialization wasn't checked. After fixing this, we still have 2 event listeners / dropdown. It looks like, `dropdowns.selectpicker()` creates a second dropdown element for each existing one (when I comment `dropdowns.selectpicker()` out, the loop iterates through the right amount of elements, when not, it is 2x more). Does anybody have any idea, why?

### DIFF
--- a/resources/gui/js/apis/gui.js
+++ b/resources/gui/js/apis/gui.js
@@ -63,7 +63,7 @@ class GuiApi extends IApi {
       if (typeof dropdown.dataset.initialized === 'undefined') {
         dropdown.addEventListener('change', eventListener);
 
-        input.dataset.initialized = 'true';
+        dropdown.dataset.initialized = 'true';
       }
     });
   }
--- a/resources/gui/js/apis/gui.js
+++ b/resources/gui/js/apis/gui.js
@@ -59,7 +59,7 @@ class GuiApi extends IApi {
       }
     };
 
-    document.querySelectorAll('.simple-value-dropdown').forEach((dropdown) => {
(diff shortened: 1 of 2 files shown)


Schneegans:
Thanks for the fix! Maybe you can do a `console.log(dropdown)` inside of the loop and check with the CEF Remote Debugging Tools for which elements it's called? Maybe you can even differentiate between the dropdowns and only add the listener to the correct one.

gmate321:
> Thanks for the fix! Maybe you can do a `console.log(dropdown)` inside of the loop and check with the CEF Remote Debugging Tools for which elements it's called? Maybe you can even differentiate between the dropdowns and only add the listener to the correct one.

After calling `dropdowns.selectpicker()` a dropdown looks like this:
`<div class="dropdown bootstrap-select simple-value-dropdown"> `
      `<select data-callback="..." class="simple-value-dropdown" ... > ... </select>`
      `<button ...>...</button>`
      `<div class="dropdown-menu " ... > ... </div>`
`</div>`

When calling `document.querySelectorAll('simple-value-dropdown')`, both the `<div>` and the `<select>` element was selected.

I think `selectpicker()` should only be called once on an element. When called multiple times it creates multiple `<div>`s, `<button>`s around the `<select>` element. I added a not selector to avoid calling  `selectpicker()` on elements where it was already called. Alternatively we could pass the plugin name to `initInputs()` and then check it with the `data-callback` name of the dropdowns.

Schneegans:
Thank you!

PR #152, opened 2020-05-01
Capture Depth Images

Schneegans:
This adds the following features:

* Save and load the scene from and to a json string (instead of a json file).
* A `/save` and a `/load` endpoint for `csp-web-api` which allows, well, saving and loading of the scene via http.
* A `&depth=true|false` parameter to the /capture endpoint of `csp-web-api` to allow capturing of 32 bit tiff depth images.

I forgot to create a feature branch for `csp-web-api`, here's the diff: https://github.com/cosmoscout/csp-web-api/compare/259f00b28e26009bb6f7ffc78ece7fbeb7ca3b89...develop

### DIFF
--- a/src/cs-core/Settings.cpp
+++ b/src/cs-core/Settings.cpp
@@ -286,7 +286,7 @@ utils::Signal<> const& Settings::onSave() const {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-void Settings::read(std::string const& fileName) {
+void Settings::loadFromFile(std::string const& fileName) {
   std::ifstream i(fileName);
 
   if (!i) {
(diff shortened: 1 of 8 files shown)


src/cs-core/Settings.cpp
```diff
@@ -329,6 +340,21 @@ void Settings::write(std::string const& fileName) const {
...
 
+std::string Settings::saveToJson() const {
+  // Tell listeners that the settings are about to be saved.
+  mOnSave.emit();
+
+  nlohmann::json settings = *this;
+
+  // Use an indentation of two space.
+  std::ostringstream o;
+  o << std::setw(2) << settings;
```
TheJhonny007, line 351: Why do we need the `setw` here? If the string contains more than two characters it shouldn't do anything. Or am I missing something?

Schneegans: It's overloaded for nlohmann::json and specifies the indentation width.

PR #154, opened 2020-05-07
Bookmarks and more

Schneegans:
The primary goal of this pull request is to add a solid bookmark system to core CosmScout VR. While implementing this I came across several unforeseen issues... See the list of changes below!

* Plugins can create and delete bookmarks via the GuiManager's API. A bookmark can have a positions in space and / or time.
  - The location of a bookmark is defined by a SPICE anchor, an optional cartesian position and an optional rotation.
  - The time of a bookmark has an optional end parameter which makes the bookmark describe a time span rather a time point.
  - Location and time are both optional, but omitting both results in a pretty useless bookmark.
  - The Bookmark Editor has been completely redesigned. It performs quite sophisticated input checking to avoid creating invalid bookmarks. 
* All times in UTC!
  - We actually mixed local timezones and UTC in our code. Now the cs::utils::convert::time namespace makes sure that we always stay in UTC.
  - The date display in the center of the screen now shows UTC in the less German-looking YYYY-MM-DD HH:MM:SS.
  - The conversion from boost::posix_time::ptime to BDT of SPICE was wrong - we did not include leap seconds! Therefore the simulation time and visualized time were off by a few minutes. And also the position of the bookmarks on the timeline was wrong...
* Many issues in the timeline JavaScript code have been fixed.
  - The code has been simplified: From 1500 lines to 750 lines.
  - It's now possible to zoom on the timeline even if time is playing.
  - Moving the timeline will properly restore the playback speed once the drag operation is finished.
  - CSS has been cleaned as well, Bookmarks on the timeline are now almost pixel-perfectly positioned.
  - The styling of the Bookmark tool-tip has been improved.
* Movable Windows!
  - We now can easily create movable and resizable windows in our UI.
  - The Calendar has been modified to be draggable and resizable.
  - The new Bookmark Editor also uses this.
* Some CSS refactoring
  - Some of the changes above required modifications of our gui.css which in turned required changes in other places. For example the measurement tools needed to be adapted.


TheJhonny007:
I get the following error when trying to run with this branch:

```
C:\C++\cosmoscout-vr\install\windows-release\bin\cosmoscout.exe --settings=../share/config/simple_desktop.json -vistaini vista.ini
[E] cosmoscout-vr ..... Failed to read settings: While parsing property 'center' in 'location' in 'bookmarks': [json.exc
eption.out_of_range.403] key 'center' not found

Process finished with exit code 1
```

Schneegans:
There were no correct default bookmarks committed yet. You can try it now!

Schneegans:
A few thoughts:
* We could use the new draggable windows for the minimap! Maybe even in a dedicated plugin?
* I will add an optional `icon` (an image file path) to the bookmark. This will allow rewriting of `csp-fly-to-locations` using only the core bookmarks. It then can use core bookmarks for the planet selection and the locations on the planets.

TheJhonny007:
> * We could use the new draggable windows for the minimap! Maybe even in a dedicated plugin?

This sounds like a good idea.

TheJhonny007:
Movable windows can easily get lost outside of the screen. We may need to ristrict them to the window bounds.

Schneegans:
I added a simple check which prevents windows from being moved offscreen.

Schneegans:
So, `csp-fly-to-locations` uses the bookmarks now. What do you think?

TheJhonny007:
> So, `csp-fly-to-locations` uses the bookmarks now. What do you think?

Is it possible to go to time and place with one click?

Schneegans:
This is exactly what Moritz' Thesis is going to be about. As the transition would be quite funny, I would leave it for now as it is.

TheJhonny007:
I find it quite unintuitive to manually put the icons into the correct folder as an end user. Can you open a file explorer with CEF?

Schneegans:
> I find it quite unintuitive to manually put the icons into the correct folder as an end user. Can you open a file explorer with CEF?

I think it could be possible, but it would be a native window. This would lead to issues if CosmoScout is running in fullscreen mode and it will never work in a VR-context.

TheJhonny007:
You can break a bookmark by inserting html in the description or title. Example: `<script>console.log("Test")</script>`

TheJhonny007:
> > I find it quite unintuitive to manually put the icons into the correct folder as an end user. Can you open a file explorer with CEF?
> 
> I think it could be possible, but it would be a native window. This would lead to issues if CosmoScout is running in fullscreen mode and it will never work in a VR-context.

Yeah, I didn't think about that.

TheJhonny007:
I think we need to deal with links to external sites. If you provide an html link and someone clicks on it, it will break the whole user interface.

Maybe this can be in another pull request.

TheJhonny007:
But in general it is quite handy to have basic html formatting in the description. You can create quite fancy info texts. Just links and scripts are broken.

Schneegans:
I added a simple check to prevent this from the Bookmark Editor side. You can still break it by directly editing the settings file, but this is ok, I guess :smile:

resources/gui/css/timeline.css
```diff
@@ -1,86 +1,105 @@
...
+
+#timeline .vis-time-axis {
+  pointer-events: none;
+}
+
+#timeline .vis-item.vis-selected {
+  border-color: rgba(20, 20, 20, 0.5);
+  background-color: rgba(20, 20, 20, 0.5);
+
+  /* For some reason active items are translate 3 pixels to the left. We compensate for that here. */
```
TheJhonny007, line 86:
```suggestion
  /* For some reason active items are translated 3 pixels to the left. We compensate for that here. */
```

src/cosmoscout/Application.cpp
```diff
@@ -1157,15 +1171,114 @@ void Application::registerGuiCallbacks() {
...
+  mGuiManager->getGui()->registerCallback("bookmark.add",
+      "Adds a new bookmark. The parameter is a JSON string as in CosmoScout's settings.",
+      std::function([this](std::string&& jsonString) {
+        cs::core::Settings::Bookmark bookmark;
+        auto                         json = nlohmann::json::parse(jsonString);
+        json.get_to(bookmark);
+        mGuiManager->addBookmark(bookmark);
+      }));
+
+  // Show the Bookmar-Editor for the given bookmark.
```
TheJhonny007, line 1192:
```suggestion
  // Show the Bookmark-Editor for the given bookmark.
```

src/cosmoscout/Application.cpp
```diff
@@ -1157,15 +1171,114 @@ void Application::registerGuiCallbacks() {
...
   // Timeline callbacks ----------------------------------------------------------------------------
 
   // Sets the current simulation time. The argument must be a string accepted by
   // TimeControl::setTime.
   mGuiManager->getGui()->registerCallback("time.setDate",
-      "Sets the current simulation time. Format must be in the format '2002-01-20 23:59:59.000'.",
-      std::function([this](std::string&& sDate) {
-        double time = cs::utils::convert::toSpiceTime(boost::posix_time::time_from_string(sDate));
-        mTimeControl->setTime(time);
+      "Sets the current simulation time. Format must be in the format '2002-01-20T23:59:59.000Z'. "
```
TheJhonny007, line 1273:
```suggestion
      "Sets the current simulation time. Format must be: '2002-01-20T23:59:59.000Z'. "
```

src/cs-core/GuiManager.cpp
```diff
@@ -449,4 +511,31 @@ void GuiManager::setSliderValue(
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+void GuiManager::onLoad() {
+  // First clear all bookmarks. In theory this could be optimized by not-reloading identical
```
TheJhonny007, line 515:
```suggestion
  // First clear all bookmarks. In theory this could be optimized by not reloading identical
```

src/cs-core/Settings.hpp
```diff
@@ -178,23 +178,49 @@ class CS_CORE_EXPORT Settings {
     utils::Property<glm::dquat> pRotation;
   } mObserver;
 
-  /// Events to show on the timenavigation bar
-  struct Event {
+  /// Bookmarks are managed in CosmoScout's core. Plugins can create and delete bookmars via the
```
TheJhonny007, line 181:
```suggestion
  /// Bookmarks are managed in CosmoScout's core. Plugins can create and delete bookmarks via the
```

src/cs-core/Settings.hpp
```diff
@@ -178,23 +178,49 @@ class CS_CORE_EXPORT Settings {
...
+  /// GuiManager's API. A bookmark can have a positions in space and / or time. It may also describe
```
TheJhonny007, line 182:
```suggestion
  /// GuiManager's API. A bookmark can have a position in space and / or time. It may also describe
```

src/cs-core/Settings.hpp
```diff
@@ -178,23 +178,49 @@ class CS_CORE_EXPORT Settings {
...
-    std::string                mDescription;
-    std::optional<Location>    mLocation;
+    /// The name of the bookmark is the only required field. It's not strictly required but a good
+    /// idea to keep this unique amongst the bookmarks for an anchor.
+    std::string mName;
+
+    /// This can be a longer text.
+    std::optional<std::string> mDescription;
+
+    /// This can be a name of a png file in share/ressources/icons.
```
TheJhonny007, line 209:
```suggestion
    /// This can be a name of a png file in share/resources/icons.
```

src/cs-core/SolarSystem.hpp
```diff
@@ -131,6 +131,16 @@ class CS_CORE_EXPORT SolarSystem {
   void flyObserverTo(std::string const& sCenter, std::string const& sFrame,
       glm::dvec3 const& position, glm::dquat const& rotation, double duration);
 
+  /// Gradually moves the observer's position from its current value to the given values. The
+  /// rotation will be chosen automatically to be donward-facing.
```
TheJhonny007, line 135:
```suggestion
  /// rotation will be chosen automatically to be downward-facing.
```

src/cs-utils/convert.hpp
```diff
@@ -79,16 +79,51 @@ CS_UTILS_EXPORT glm::dvec3 lngLatToNormal(glm::dvec2 const& lngLat, double radiu
...
+/// Time in CosmoScout VR is passed around in different formats.
+/// * Strings usually store time in the ISO format YYYY-MM-DDTHH:MM:SS.fffZ. The 'Z' suffix is not
+///   really required on the C++ side, as time strings are always considered to be in UTC. This
+///   format is also directly convertible to JavaScript Dates, here however the 'Z' is required!
+///   Else the Date object will be in your local time zone. So it's a good practive to always append
+///   the 'Z'.
+/// * boost::posix_time::ptime is used for conversions and is also always in UTC.
+/// * SPICE time is stored in doubles representing Barycentric Dynamical Time (seconds since
+///   2000-01-01 12:00:00). This however differs from the number of UTC seconds since that date
+///   as some leap seconds are taken into account. The conversion methods below take this into
```
TheJhonny007, line 91: This sentence is confusing.

JonasGilg (changes requested):

JonasGilg (approved):

src/cs-utils/convert.hpp
Schneegans: I rephrased it. Do you think it's better now?

src/cs-utils/convert.hpp
```diff
@@ -83,13 +83,12 @@ CS_UTILS_EXPORT glm::dvec2 normalToLngLat(glm::dvec3 const& normal, double radiu
...
+///   Else the Date object will be in your local time zone. So it's a good practice to always append
 ///   the 'Z'.
 /// * boost::posix_time::ptime is used for conversions and is also always in UTC.
-/// * SPICE time is stored in doubles representing Barycentric Dynamical Time (seconds since
-///   2000-01-01 12:00:00). This however differs from the number of UTC seconds since that date
-///   as some leap seconds are taken into account. The conversion methods below take this into
-///   account.
+/// * SPICE time is stored in doubles representing Barycentric Dynamical Time (BDT, seconds since
+///   2000-01-01 12:00:00). Note that this is not the same as UTC seconds since 2000-01-01 12:00:00
+///   because BDT considers leap seconds. The conversion methods below take this into account.
```
TheJhonny007, line 91: Time system abbreviations are never what you expect 😆.

```suggestion
/// * SPICE time is stored in doubles representing Barycentric Dynamical Time (TDB, seconds since
///   2000-01-01 12:00:00). Note that this is not the same as UTC seconds since 2000-01-01 12:00:00
///   because TDB considers leap seconds. The conversion methods below take this into account.
```

JonasGilg (changes requested):

src/cs-utils/convert.hpp
Schneegans: :open_mouth:

JonasGilg (approved):

src/cs-utils/convert.hpp
TheJhonny007: > 😮

We have to thank the hatred between France and England for some of those apparently. They couldn't agree on the English or French abbreviation, so they ordered the letters so no one got their will 🤣

Schneegans:
Thank you for the review!

PR #155, opened 2020-05-09
csp-custom-web-ui

Schneegans:
This includes csp-custom-web-ui as a default plugin.


Schneegans:
The scrolling on the DLR page works for me after the "fix" i just committed. Can you confirm this? The Spotify-Issue will be fixed by the incoming `feature/minimap` pull request. (due to the changes in https://github.com/cosmoscout/cosmoscout-vr/blob/feature/minimap/src/cs-gui/internal/LifeSpanHandler.cpp#L16)

TheJhonny007:
You can merge, once the submodule is updated.

PR #156, opened 2020-05-10
Add possibiltity to auto-hide window headers

Schneegans:
With the class `.auto-hide-header` window title bars can now be hidden automatically. This reduces screen clutter and is currently used for the calendar and will be used for the minimap soon. The title bar is revealed when the pointer is above the window.

In `cosmoscout.html` actually only four lines changed because a `<div class="window-wrapper">` was added. The rest is indentation. :open_mouth:

### DIFF
--- a/resources/gui/cosmoscout.html
+++ b/resources/gui/cosmoscout.html
@@ -632,7 +632,7 @@
                 id="bookmark-editor-frame">
               <div class="input-group-append">
                 <button class="btn glass" type="button" data-toggle="tooltip"
-                  title="Use current Planet"><i class="material-icons">place</i></button>
+                  title="Use current Body"><i class="material-icons">place</i></button>
               </div>
               <div class="invalid-feedback"></div>
             </div>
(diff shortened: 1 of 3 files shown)


resources/gui/cosmoscout.html
```diff
@@ -499,188 +499,194 @@
...
+
+            <div class="col-12 input-group">
+              <input type="text" placeholder="Center" class="form-control"
+                id="bookmark-editor-center">
+              <input type="text" placeholder="Frame" class="form-control"
+                id="bookmark-editor-frame">
               <div class="input-group-append">
                 <button class="btn glass" type="button" data-toggle="tooltip"
-                  title="Use current time"><i class="material-icons">update</i></button>
+                  title="Use current Planet"><i class="material-icons">place</i></button>
```
TheJhonny007, line 635:
```suggestion
                  title="Use current Body"><i class="material-icons">place</i></button>
```

PR #158, opened 2020-05-13
:wrench: make_externals.bat now works with Ninja as a generator.

TheJhonny007:
This pull request aims to improve build times on windows by making the `make_externals.bat` compatible with the Ninja generator.

Here are some measurements on build times:
```
.\make_externals.bat -GNinja -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe
Ninja Clean: 06:08 (mm:ss)
Ninja NOOP:  02:07 (mm:ss)

----------------------------------------------------------------------------------

.\make_externals.bat -G "Visual Studio 16 2019" -A x64
MSBuild Clean: 10:52 (mm:ss)
MSBuild NOOP:  02:44 (mm:ss)
```

Both compilations were done from the `x64 Native Tools Command Prompt for VS 2019`.

I will also add some documentation. Before this is ready to merge.

### DIFF
--- a/make_externals.bat
+++ b/make_externals.bat
@@ -9,13 +9,14 @@ rem ----------------------------------------------------------------------------
 rem ---------------------------------------------------------------------------------------------- #
 rem Make sure to run "git submodule update --init" before executing this script!                   #
 rem Default build mode is release, if "set COSMOSCOUT_DEBUG_BUILD=true" is executed before, all    #
-rem dependecies will be built in debug mode.                                                       #
+rem dependencies will be built in debug mode.                                                      #
 rem Usage:                                                                                         #
 rem    make_externals.bat [additional CMake flags, defaults to -G "Visual Studio 15 Win64"]        #
 rem Examples:                                                                                      #
 rem    make_externals.bat                                                                          #
 rem    make_externals.bat -G "Visual Studio 15 Win64"                                              #
 rem    make_externals.bat -G "Visual Studio 16 2019" -A x64                                        #
+rem    make_externals.bat -GNinja -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe            #
 rem ---------------------------------------------------------------------------------------------- #
 
 rem The CMake generator and other flags can be passed as parameters.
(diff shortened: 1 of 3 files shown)


Schneegans:
Should this also work for `make.bat` already? CMake fails to find boost when I use Ninja. Do you have an idea why this could be?

TheJhonny007:
> Should this also work for `make.bat` already?

Yes, I used it with `make.bat` for months already. Only the externals needed a tweak.

>  CMake fails to find boost when I use Ninja. Do you have an idea why this could be?

How do you normally provide the BOOST_ROOT? I didn't have to change anything to my workflow for the `make.bat`. I normally provide it through CLions target specific environment variables.

Schneegans:
Just with `set BOOST_ROOT=C:\local\boost_1_69_0` before the `make.bat`.

TheJhonny007:
Ok, here is how I compile it with pure command line:

1. I start **x64 Native Tools Command Prompt for VS 2019**

2. I enter the following commands:
```bash
set BOOST_ROOT=C:\C++Libs\boost\msvc-142\boost_1_71_0
.\make.bat -GNinja -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe
```

TheJhonny007:
I think I fixed it. It required the `CMAKE_BUILD_TYPE` variable to be set at configure time.

TheJhonny007:
Do you think we should run Windows CI with Ninja to get faster builds or should we avoid that, so CI ensures that MSBuid always works?

make_externals.bat
```diff
@@ -9,13 +9,14 @@ rem ----------------------------------------------------------------------------
...
 rem Default build mode is release, if "set COSMOSCOUT_DEBUG_BUILD=true" is executed before, all    #
-rem dependecies will be built in debug mode.                                                       #
+rem dependencies will be built in debug mode.                                                      #
 rem Usage:                                                                                         #
 rem    make_externals.bat [additional CMake flags, defaults to -G "Visual Studio 15 Win64"]        #
 rem Examples:                                                                                      #
 rem    make_externals.bat                                                                          #
 rem    make_externals.bat -G "Visual Studio 15 Win64"                                              #
 rem    make_externals.bat -G "Visual Studio 16 2019" -A x64                                        #
+rem    make_externals.bat -GNinja -DCMAKE_C_COMPILER=cl.exe -DCMAKE_CXX_COMPILER=cl.exe            #
```
Schneegans, line 19: You could add a similar comment to `make.bat`

PR #168, opened 2020-05-18
Allow multiples calls to CelestialObserver::moveTo per frame

Schneegans:
Even if the animation time was set to zero, it took one frame to update the observer's position and rotation. Hence multiple calls to CelestialObserver::moveTo would "overwrite" each other.

Calling something like this didn't work (but does now):

```javascript
CosmoScout.callbacks.navigation.setPosition(x, y, z, 0);
CosmoScout.callbacks.navigation.setRotation(x, y, z, w, 0);
```


Schneegans:
Is the **Hide whitespace changes** option in the Github diff new? I've never seen this before.... Anyways, it's very helpful for this PR!

PR #169, opened 2020-05-18
Add the possibilty to capture images in multiple formats

Schneegans:
This adds a `&format=png` (or `tiff` or `jpeg`) parameter to the `/capture` endpoint of `csp-web-api` which lets you capture depth and color images in different formats.

The main advantage is speed, since png or tiff images tend to be quite large.

### DIFF
--- a/plugins/csp-web-api/src/Plugin.cpp
+++ b/plugins/csp-web-api/src/Plugin.cpp
@@ -49,13 +49,52 @@ namespace {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// Converts a void* to a std::vector<std::byte> (which is given through a void* as well). So this is
-// pretty unsafe, but I think it's the only way to make stb_image write to a std::vector<std::byte>.
-void pngWriteToVector(void* context, void* data, int len) {
+// Appends void* data to a std::vector<std::byte> (which is given through a void* as well). So this
+// is pretty unsafe, but I think it's the only way to make stb_image write to a
+// std::vector<std::byte>.
+void stbWriteToVector(void* context, void* data, int len) {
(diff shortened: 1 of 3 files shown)


plugins/csp-web-api/src/Plugin.cpp
```diff
@@ -345,73 +386,81 @@ void Plugin::update() {
...
-          mCaptureWidth, mCaptureHeight, mCaptureGui);
+      logger().debug("Capturing capture for /capture request: resolution = {}x{}, show gui = {}, "
+                     "depth = {}, format = {}",
+          mCaptureWidth, mCaptureHeight, mCaptureGui, mCaptureDepth, mCaptureFormat);
 
       auto* window = GetVistaSystem()->GetDisplayManager()->GetWindows().begin()->second;
       window->GetWindowProperties()->GetSize(mCaptureWidth, mCaptureHeight);
 
+      // We encode image data in the main thread as this is not thread-safe.
+      stbi_flip_vertically_on_write(1);
```
TheJhonny007, line 397: Is it possible to cache the value of this flag? So we can reset it to the previous value, even if it wasn't 0?

Schneegans: Well, there is no "official" getter. I think we could just access the global `static int stbi__flip_vertically_on_write` but this is not intended, I guess. And it may be even impossible with MSVC as it's not exported. So we should simply reset it to its default state. Or do you see a better way?

TheJhonny007: Yeah, I looked at the definition of the variable. As a **global** `static` in a header file, we don't even need to bother to reset it at all, since each `.cpp` file has it's own version. So we can leave it as is.

PR #170, opened 2020-05-19
Geo-Coding

### LINKED ISSUE #80: Re-enable the location search bar
This should be now in the csp-fly-to-locations plugin.

Schneegans:
This fixes #80 . It adds a new javascript API which can be used to perform forward and reverse geocoding. Here are some examples:

```javascript
CosmoScout.geocode.forward("earth", "berlin", (l) => console.log(l.name, l.diameter))
CosmoScout.geocode.forward("mars", "gale", (l) => console.log(l.longitude, l.latitude))
```

```javascript
CosmoScout.geocode.reverse("earth", 5, 48, (l) => console.log(l))
CosmoScout.geocode.reverse("mars", 137.81, -5.37, (l) => console.log(l))
```

The geo-coding for Earth uses OpenStreetMap, all other planets use CSV files obtained from https://planetarynames.wr.usgs.gov/AdvancedSearch.

There is now a search bar beneath the timeline which supports queries like
* **"berlin"** Fly to something called like "Berlin" on the current planet.
* **"venus:"** Fly to Venus
* **"mars:olymps mns"** Fly to something called like "olymps mns" on Mars.

The flag tool now uses the reverse geo-coding API to show its location.


TheJhonny007:
When you are on Mars, and type in `Earth` you are traveling to `Mars:Barth`.

resources/gui/js/apis/geocode.js
```diff
@@ -0,0 +1,230 @@
...
+   * @param {number} latitude  In degrees.
+   * @private
+   */
+  _toCartesian(longitude, latitude) {
+    var lon = longitude * Math.PI / 180.0;
+    var lat = latitude * Math.PI / 180.0;
+
+    var x = Math.cos(lat) * Math.sin(lon);
+    var y = Math.sin(lat);
+    var z = Math.cos(lat) * Math.cos(lon);
```
TheJhonny007, line 214:
```suggestion
    const lon = longitude * Math.PI / 180.0;
    const lat = latitude * Math.PI / 180.0;

    const x = Math.cos(lat) * Math.sin(lon);
    const y = Math.sin(lat);
    const z = Math.cos(lat) * Math.cos(lon);
```

Schneegans:
> When you are on Mars, and type in `Earth` you are traveling to `Mars:Barth`.
> 
> **Edit:** Ah, I see you need to put a colon after planet names. This is a little bit unintuitive, especially if you fly to locations that don't resemble the name you put in at all, like on Mars -> `venus` travels to `Mars:Vernal`.

Do you have an suggestion how this could be improved? The only thing I could think of is using two input fields next to each other, one for the planet and one for the location. But this would be a bit slower to work with.

TheJhonny007:
> > When you are on Mars, and type in `Earth` you are traveling to `Mars:Barth`.
> > **Edit:** Ah, I see you need to put a colon after planet names. This is a little bit unintuitive, especially if you fly to locations that don't resemble the name you put in at all, like on Mars -> `venus` travels to `Mars:Vernal`.
> 
> Do you have an suggestion how this could be improved? The only thing I could think of is using two input fields next to each other, one for the planet and one for the location. But this would be a bit slower to work with.

Maybe change the priority. For example if no colon is present assume first that it is a planet and if no matching is found search for a location on the current planet.

I mean how often are you searching for a place called Mars, Neptune, Venus or Earth?

Schneegans:
The difficulty is that we currently do not know on the JavaScript side which anchors exist in the scene. So it would be rather complex to implement - and how often will someone type in the target's name if it is possible to click on a bookmark or on the anchor label instead?

PR #174, opened 2020-05-28
Add `sourceLocation` hint to all core js files.

TheJhonny007:
This makes it possible to debug javascript properly.


Schneegans:
Can you explain what you mean by "properly"? As all these files are included via the `script` tags, the Chromium Developper Tools know where to find the files...?

TheJhonny007:
> Can you explain what you mean by "properly"? As all these files are included via the `script` tags, the Chromium Developer Tools know where to find the files...?

Oh yeah, I probably was too tired to realize that not all JS was included dynamically 🤣

PR #181, opened 2020-06-15
Recorder

Schneegans:
This adds a simple recorder plugin which allows basic recording of high-quality videos. See the Readme (https://github.com/cosmoscout/cosmoscout-vr/tree/feature/recorder/plugins/csp-recorder) for usage instructions!

It also adds some required changes to `csp-web-api` (allowing default values for resolution and show-gui parameters), adds a new `CosmoScout.callbacks.navigation.setBodyFull()` callback and tweaks the `cs-core` handling of frame transitions. The observer center and frame members are now updated also during animations - this was required for the recording of animations.


plugins/csp-recorder/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Usage
+
+The capturing is done in three steps:
+1. **Record Something:** Setup a scene with as little plugins as possible as frame rates matter.
+Maybe it's a good idea to only render simple planets.
+The hit the record button beneath the timeline and fly around.
+When finished, hit the record button once more.
+2. **Capture the Frames:** Step 1 has produced a python script called `recording-<current date>.py`.
```
bernstein, line 35: I would mention where the script is saved to, maybe something like this
```suggestion
2. **Capture the Frames:** Step 1 has produced a python script called `recording-<current date>.py` next to the cosmoscout executable.
```

plugins/csp-recorder/src/Plugin.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include "../../../src/cs-core/Settings.hpp"
+#include "../../../src/cs-utils/Property.hpp"
+
+#include <fstream>
+
+namespace csp::recorder {
+
+/// This plugin allows basic capturing of high-quality videos using csp-web-api. 'Basic' means that
+/// (for now) only the observer transformation, the simulation time and the exposure of the HDR mode
+/// is captured. This however, can be changed in future.
```
bernstein, line 20:
```suggestion
/// is captured. This however, can be changed in the future.
```

plugins/csp-recorder/src/Plugin.cpp
```diff
@@ -0,0 +1,238 @@
...
+    // the current recording session. So open the file!
+    if (!mOutFile.is_open()) {
+
+      // Reset the frame counter.
+      mFrameCounter = 0;
+
+      // We use the current date as a filename.
+      std::string fileName =
+          "recording-" +
+          cs::utils::convert::time::toString(boost::posix_time::microsec_clock::local_time()) +
```
bernstein, line 127: the plugin did not create the script on my windows machine. It has some issue with the time string because changing the filename to "recording-bla.py" just for testing worked.

plugins/csp-recorder/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+
+if (NOT CSP_RECORDER)
+  return()
+endif()
+
+# build plugin -------------------------------------------------------------------------------------
+
+file(GLOB SOURCE_FILES src/*.cpp)
+
+# Resoucre files and header files are only added in order to make them available in your IDE.
```
bernstein, line 17:
```suggestion
# Resource files and header files are only added in order to make them available in your IDE.
```

plugins/csp-recorder/src/Plugin.cpp
bernstein: Adding `cs::utils::replaceString(fileName, ":", "-");` fixed it.

plugins/csp-recorder/src/Plugin.cpp
```diff
@@ -0,0 +1,238 @@
...
+          ".py";
+
```
bernstein, line 129:
```suggestion
      cs::utils::replaceString(fileName, ":", "-");

```

PR #190, opened 2020-07-21
[Vestec] Merge develop

octfx:
Merge the current `develop` branch into `feature/vestec`.

See https://github.com/cosmoscout/csp-vestec/pull/2


Flatken:
Would be good if you can run clang format on the code

octfx:
> Would be good if you can run clang format on the code
Done


> Hi Hannes. I did some further commits on your branch and tested things! I looks very good thank you. One thing you already mentioned we should add is deleting nodes with esc

The `del` key event now gets injected into the main GUI. 53b15f2dbe906d34534de608ec6573bf81924495

PR #192, opened 2020-08-25
Faster builds

JonasGilg:
I replaced the `Invoke-WebRequest` commands on windows with `curl.exe` commands, which reduces download times by a few minutes.

Unzipping of CSpice now only extracts the relevant folders.


Schneegans (approved):
Looks good! I did not know that curl is included in Windows 10. I am not sure about Windows 7, but tthis shouldn't be used anymore anyways...

PR #196, opened 2020-09-07
Initial changelog draft for v1.3.0

Schneegans:
I started drafting the changelog for 1.3.0. It's already quite a lot! What do you think? Did I miss something important?


JonasGilg (changes requested):
We also fixed Vive support and moved all plugins into the core repository.

Schneegans:
I added a note regarding the Vive driver. The plugin move was already mentioned.

> All default **plugins are now part of the source tree** and no individual submodules anymore. This simplifies the software development cycle significantly.

Issue #102, opened 2020-01-02
CosmoScout crashes on MROs SPICE frame ending


octfx:
1. Start CosmoScout with full config
2. Travel to Mars
3. Set time to `12.03.2018 12:00:00`
4. Set playback time to any positive value
5. Wait until spice frame ends

## Expected Behavior
No crash to desktop.

Schneegans:
Thanks for the report!

PR #198, opened 2020-09-08
Fix SPICE crash

Schneegans:
This fixes issue #102 by not attempting to draw a trajectory when the target body is not in existence.

This PR also updates Vista to the current develop branch.

### DIFF
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -112,7 +112,7 @@ Last but not least, the far clipping plane depends on the scene scale: Near clip
     ...
   }
   ```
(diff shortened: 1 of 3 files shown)


JonasGilg:
This bug is only partially fixed. When following MRO while its spice frame expires the anchor labels throw errors or the program crashes with `double free or corruption (out)`.

Schneegans:
This is due to the offset between TDB and UTC. You must set the `endExistence` to something like `"2018-03-12 14:29:00.000"`.

JonasGilg (approved):

PR #200, opened 2020-09-09
Console fixes

Schneegans:
This makes the JavaScript console auto-complete to the longest shared prefix if there are multiple matches. It also makes sure that the suggestions are shown above the sidebar, making them always clickable.


JonasGilg (approved):
Works like a charm 😄

Schneegans:
Yay! Pull request 200!

PR #202, opened 2020-09-14
Catch most SPICE errors

Schneegans:
This fixes several SPICE-related crashes.

This fixes issue #102

### DIFF
--- a/plugins/csp-anchor-labels/src/AnchorLabel.cpp
+++ b/plugins/csp-anchor-labels/src/AnchorLabel.cpp
@@ -6,6 +6,8 @@
 
 #include "AnchorLabel.hpp"
 
+#include "logger.hpp"
+
 #include "../../../src/cs-core/GuiManager.hpp"
 #include "../../../src/cs-core/InputManager.hpp"
 #include "../../../src/cs-core/SolarSystem.hpp"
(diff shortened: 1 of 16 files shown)


plugins/csp-anchor-labels/src/AnchorLabel.cpp
```diff
@@ -104,34 +106,40 @@ void AnchorLabel::update() {
...
 
-    auto rot = glm::toQuat(glm::dmat3(x, y, z));
-    mAnchor->setAnchorRotation(rot);
+      mAnchor->update(simulationTime, mSolarSystem->getObserver());
 
-    mAnchor->update(simulationTime, mSolarSystem->getObserver());
+    } catch (std::exception const& e) {
+      // Getting the relative transformation may fail due to insufficient SPICE data.
+      logger().warn("AnchorLabel::update failed for '{}': {}", mBody->getCenterName(), e.what());
+    }
```
JonasGilg, line 142: While this prevents the application from a crash, it leaves it in an almost unusable state. When you follow MRO until its frame expires no more movement is possible. The only recovery possible is to revert time to a point where MRO existed and then move out of frame.

A better solution would be to change the center and frame once this happens.

JonasGilg (changes requested):

Schneegans:
Soo... can you test this once more? Simply changing the SPICE frame is obviously impossible as we do not have enough data to compute the position / orientation of it in the first place. I have now implmented some kind of workaround which keeps track of the last known working simulation time and uses this in case of an error to reset the observer frame. This seems to work but feels kind of fragile. At least, the application is still usable if the current SPICE frame fails to update....

src/cs-core/SolarSystem.cpp
```diff
@@ -89,6 +89,25 @@ scene::CelestialObserver const& SolarSystem::getObserver() const {
...
+  try {
+    mObserver.getRelativePosition(mTimeControl->pSimulationTime.get(),
+        scene::CelestialAnchor("Solar System Barycenter", "J2000"));
+  } catch (...) {
+    // In case of an error, we reset the observer. This can throw an error itself, but we cannot do
+    // anything if that happens.
+    try {
+      mObserver.changeOrigin("Solar System Barycenter", "J2000", lastWorkingSimulationTime);
+    } catch (std::exception const& e) {
+      logger().warn("Failed to reset the Observer SPICE frame: {}", e.what());
```
JonasGilg, line 104: If this happens, shouldn't that be an error or critical error, since the application state is not recoverable?

JonasGilg (changes requested):
While it might be hacky, it is much better than having a broken app state. From a user perspective it actually feels like intended behaviour.

src/cs-core/SolarSystem.cpp
Schneegans: Good point!

JonasGilg (approved):

PR #208, opened 2020-09-18
Apply clang-tidy

Schneegans:
It still reports many additional warnings, but these are the ones which could be applied automatically and semi-automatically.


plugins/csp-measurement-tools/src/voronoi/Breakpoint.hpp
```diff
@@ -24,15 +24,15 @@ class Breakpoint {
   Vector2f const& position() const;
   Edge            finishEdge(Vector2f const& end) const;
 
-  Arc *       mLeftArc, *mRightArc;
-  Breakpoint *mLeftChild, *mRightChild, *mParent;
+  Arc *       mLeftArc{nullptr}, *mRightArc{nullptr};
+  Breakpoint *mLeftChild{nullptr}, *mRightChild{nullptr}, *mParent{nullptr};
```
JonasGilg, line 28: This is a rule we enabled, but clang-tidy can't apply it to member variables due to technical limitations. Especially with the use of pointer we should enforce that.
```suggestion
  Arc* mLeftArc{nullptr};
  Arc* mRightArc{nullptr};
  
  Breakpoint* mLeftChild{nullptr};
  Breakpoint* mRightChild{nullptr};
  Breakpoint* mParent{nullptr};
```

plugins/csp-measurement-tools/src/voronoi/VoronoiGenerator.hpp
```diff
@@ -49,8 +49,8 @@ class VoronoiGenerator {
   void finishEdges();
 
   Beachline mBeachline;
-  double    mSweepline;
-  double    mMaxY, mMinY;
+  double    mSweepline{0.0};
+  double    mMaxY{0.0}, mMinY{0.0};
```
JonasGilg, line 53:
```suggestion
  double    mMaxY{0.0};
  double    mMinY{0.0};
```

src/cs-graphics/internal/tiny_gltf_helper.hpp
```diff
@@ -18,61 +18,74 @@ namespace cs::graphics::internal {
...
-  } else if (accessor.type == TINYGLTF_TYPE_MAT3) {
+  }
+  if (accessor.type == TINYGLTF_TYPE_MAT3) {
     return 9;
-    //} else if (accessor.type == tinygltf::TINYGLTF_TYPE_MAT4) { return 16;
-  } else {
-    return 16;
   }
+  // if (accessor.type == tinygltf::TINYGLTF_TYPE_MAT4) { return 16;
+  return 16;
```
JonasGilg, line 38:
```suggestion
  }
  // accessor.type == tinygltf::TINYGLTF_TYPE_MAT4
  return 16;
```

JonasGilg (changes requested):

JonasGilg:
> It still reports many additional warnings, but these are the ones which could be applied automatically and semi-automatically.

But no errors reported, right? Because all relevant warnings are errors. Most of the warnings are warnings, because there are instances which couldn't be fixed due to third party libraries.

Schneegans:
I fixed many but not all warnings. I think if it builds fine in CI, it's already a good step forward nonetheless.

plugins/csp-trajectories/src/Plugin.hpp
```diff
@@ -34,7 +34,8 @@ class Plugin : public cs::core::PluginBase {
 
         /// Without this, some versions of clang claim this type to be non-default-constructible...
         /// Also, Trail() = default; does not work...
-        Trail(){};
+        Trail() = default;
+        ;
```
JonasGilg: Either the comment or the code is wrong now.

JonasGilg (changes requested):

plugins/csp-trajectories/src/Plugin.hpp
Schneegans: This was "auto-fixed" by clang-tidy. Now it should work again... MSVC still complains. I'll see about that on Monday.

JonasGilg (approved):

PR #213, opened 2020-11-20
Feature/additional colormap features

### LINKED ISSUE #212: Additional features for ColorMaps
## The Problem
For my plugin [csp-volume-rendering](https://github.com/cosmoscout/csp-volume-rendering) I use the ColorMap class to deserialize transfer functions.
However, there are some additional features I needed in that class:

* Deserialize json from strings instead of files
* Access the created ColorMap as a vector of colors instead of as a texture

## The Solution
I added the ability to load ColorMaps from strings using a second constructor.
The vector of colors is saved during construction and can be accessed using the getRawData() method.

## The Alternatives
The constructor could also be implemented as one constructor with a flag, that specifies whether to handle a given string as a path or as a json string.
If this flag had a default value of "path" the change would not be breaking any previous use of the ColorMap class.

TheFridge76:
Resolves #212

### DIFF
--- a/plugins/csp-lod-bodies/src/PlanetShader.cpp
+++ b/plugins/csp-lod-bodies/src/PlanetShader.cpp
@@ -84,7 +84,7 @@ PlanetShader::PlanetShader(std::shared_ptr<cs::core::Settings> settings,
         name.erase(0, lastSlashIdx + 1);
       }
 
-      mColorMaps.insert(std::make_pair(name, cs::graphics::ColorMap(file)));
+      mColorMaps.insert(std::make_pair(name, cs::graphics::ColorMap(std::filesystem::path(file))));
       pGuiManager->getGui()->callJavascript(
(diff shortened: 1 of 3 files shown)


Schneegans:
Looks good to me. I will try to fix the CI build in `develop`, then we can test whether the checks pass for this PR!

Schneegans:
Could you please merge `develop` and see whether the checks succeed?

TheFridge76:
gcc 7.5.0 apparently doesn't support std::filesystem yet. Could we switch to a newer version of gcc or should I just use boost::filesystem for now?

Schneegans:
I think you should use `boost::filesystem` for now, as we use it in other places as well. Then, in the near future, we could update all occurrences. Updating the minimum required gcc version will introduce some other (documentation-related) changes which are not in the scope of this PR.

Schneegans (approved):

Schneegans:
Thank you once more :smile: !

PR #219, opened 2020-12-09
[Vestec] Transfer function editor

TheFridge76:
Adds an API for creating transfer function editors
As this branch is based on `develop`, #217 should be merged first.


Flatken:
Thank you

PR #223, opened 2021-01-04
Faster loading of stars on initial startup

JonasGilg:
Replaced the string split method in utils with a faster one. The stars plugin now makes use of it. It reduced inital loading times from 20 to 15 seconds.


Schneegans (approved):
Looks good! Thank you!

PR #225, opened 2021-01-05
Start supporting unity builds and pch in CosmoScout itself.

### LINKED ISSUE #133: Precompiled Headers
CMake introduced precompiled headers in 3.16 (https://cmake.org/cmake/help/v3.16/command/target_precompile_headers.html#command:target_precompile_headers), which makes it possible to easily create them platform independently.

I think we should explore that to improve compile times (especially on Windows).

JonasGilg:
This will finally resolve #133.

Tests  | Windows | Linux (WSL)
-- | -- | --
NONE | 03:28 |  01:56
UNITY | 01:30 |  01:40
PCH | 01:40 |  01:57
UNITY+PCH | 01:13 |  01:50


Schneegans:
Great work! At least in some situations this seems to bring a significant speedup! I did some timings on my Linux machine (gcc 10) and they come quite close to your WSL timings:

 | | Fresh build | convert.cpp changed | convert.hpp changed
-- | -- | -- | --
NONE | 01:50 | 00:10 | 00:46
UNITY | 01:42 | 00:14 | 01:05
PCH | 01:49 | 00:06 | 00:57
UNITY+PCH | 01:55 | 00:14 | 01:10

What do you think: Should we change the default values to `false`? Given these numbers I would rather disable them on Linux.

Another thing I would suggest is to add a comment to the beginning of the `*.pch` files like this:

```cpp
//                               This file is part of CosmoScout VR                               //
//      and may be used under the terms of the MIT license. See the LICENSE file for details.     //
//                        Copyright: (c) 2019 German Aerospace Center (DLR)                       //

// Rationale: Which headers should be put in here?

```

JonasGilg:
> Great work! At least in some situations this seems to bring a significant speedup! I did some timings on my Linux machine (gcc 10) and they come quite close to your WSL timings:

Ah great, I just wanted to start testing it on a real Ubuntu machine, but you have spared me the time :)

> What do you think: Should we change the default values to `false`? Given these numbers I would rather disable them on Linux.

Yes we should do that.

> Another thing I would suggest is to add a comment to the beginning of the `*.pch` files like this:
> 
> ```c++
> ////////////////////////////////////////////////////////////////////////////////////////////////////
> //                               This file is part of CosmoScout VR                               //
> //      and may be used under the terms of the MIT license. See the LICENSE file for details.     //
> //                        Copyright: (c) 2019 German Aerospace Center (DLR)                       //
> ////////////////////////////////////////////////////////////////////////////////////////////////////
> 
> // Rationale: Which headers should be put in here?
> ```

I can do that. My current rationale was a little by feeling. I think for smaller modules I included all headers which were used at least 3 times and for larger modules I used headers which were used at least 4 times, but it is hard to find a good guideline, since transient includes can basically include anything many times.

Schneegans:
Have you tested only including external headers? I think this could actually speed things in the development cycle since the pch's do not need to be recompiled...?

JonasGilg:
> Have you tested only including external headers? I think this could actually speed things up in the development cycle since the pch's do not need to be recompiled...?

I did test it, but only with clean builds. I can test it again with a modifying build. Maybe we can also split them up into an internal and an external header? I will test all these.

JonasGilg:
Tests                       | clean  | modified
---------------------|--------|----------
ext & int same file | TODO | TODO
ext only                  | TODO | TODO
internal extra file   | TODO | TODO

JonasGilg:
I think we can't get much better than that. Windows clean build times have been reduced by 66% and for Linux it is disabled by default.

Reusing precompiled headers aren't viable at the moment.

make.sh
```diff
@@ -35,6 +35,18 @@ case "$COSMOSCOUT_DEBUG_BUILD" in
   (true) echo "CosmoScout VR debug build is enabled!"; BUILD_TYPE=Debug;
 esac
 
+# Check if unity build is disabled with "export COSMOSCOUT_NO_UNITY_BUILD=true".
+UNITY_BUILD=Off
+case "$COSMOSCOUT_NO_UNITY_BUILD" in
+  (false) echo "CosmoScout VR debug build is enabled!"; UNITY_BUILD=On;
```
Schneegans, line 41:
```suggestion
  (false) echo "Unity build is enabled!"; UNITY_BUILD=On;
```

make.sh
```diff
@@ -35,6 +35,18 @@ case "$COSMOSCOUT_DEBUG_BUILD" in
...
+esac
+
+# Check if precompield headers should not be used with "export COSMOSCOUT_NO_PCH=true".
+PRECOMPILED_HEADERS=Off
+case "$COSMOSCOUT_NO_PCH" in
+  (false) echo "CosmoScout VR debug build is enabled!"; PRECOMPILED_HEADERS=On;
```
Schneegans, line 47:
```suggestion
  (false) echo "Precompiled headers are enabled!"; PRECOMPILED_HEADERS=On;
```

make.sh
```diff
@@ -35,6 +35,18 @@ case "$COSMOSCOUT_DEBUG_BUILD" in
   (true) echo "CosmoScout VR debug build is enabled!"; BUILD_TYPE=Debug;
 esac
 
+# Check if unity build is disabled with "export COSMOSCOUT_NO_UNITY_BUILD=true".
+UNITY_BUILD=Off
+case "$COSMOSCOUT_NO_UNITY_BUILD" in
+  (false) echo "CosmoScout VR debug build is enabled!"; UNITY_BUILD=On;
+esac
+
+# Check if precompield headers should not be used with "export COSMOSCOUT_NO_PCH=true".
```
Schneegans, line 44:
```suggestion
# Check if precompiled headers should be used with "export COSMOSCOUT_NO_PCH=false".
```

make.sh
```diff
@@ -35,6 +35,18 @@ case "$COSMOSCOUT_DEBUG_BUILD" in
   (true) echo "CosmoScout VR debug build is enabled!"; BUILD_TYPE=Debug;
 esac
 
+# Check if unity build is disabled with "export COSMOSCOUT_NO_UNITY_BUILD=true".
```
Schneegans, line 38:
```suggestion
# Check if unity build is enabled with "export COSMOSCOUT_NO_UNITY_BUILD=false".
```

Schneegans (changes requested):
There are some errors in the `*.sh` files due to the changed default values. Actually, we also should update https://github.com/cosmoscout/cosmoscout-vr/blob/feature/faster-builds/docs/install.md#getting-the-dependencies in order to reflect these changes.

Looking at this documentation file, it's pretty confusing to have different defaults for all those scripts... Should we consider changing it this way:
* Externals are built with PCH and Unity builds per default.
* CosmoSout VR is built without them per default.
* CI uses PCH and unity builds in order to catch any related errors.
* Also I think that we could improve the names of the env variables to remove the doulbe-negation (no-pch = false). We could rename them consistently to `COSMOSCOUT_USE_UNITY_BUILD` and `COSMOSCOUT_USE_PCH`

What do you think?

JonasGilg:
> There are some errors in the `*.sh` files due to the changed default values. Actually, we also should update https://github.com/cosmoscout/cosmoscout-vr/blob/feature/faster-builds/docs/install.md#getting-the-dependencies in order to reflect these changes.
> 
> Looking at this documentation file, it's pretty confusing to have different defaults for all those scripts... Should we consider changing it this way:
> 
> * Externals are built with PCH and Unity builds per default.
> * CosmoSout VR is built without them per default.
> * CI uses PCH and unity builds in order to catch any related errors.
> * Also I think that we could improve the names of the env variables to remove the doulbe-negation (no-pch = false). We could rename them consistently to `COSMOSCOUT_USE_UNITY_BUILD` and `COSMOSCOUT_USE_PCH`
> 
> What do you think?

I agree with everything you say, except I would prefer if the Windows build would use them per default. If you change a frequently used header you still save over a minute in compilation time, which adds up quite a lot.

Schneegans:
This is also fine for me, but then you have to duplicate the table with the default values several times. But this is OK, I guess :smile:

Schneegans (approved):
Thank you! Looks great!

PR #226, opened 2021-01-06
Update doctest to 2.4.4.

JonasGilg:
This updates doctest to version 2.4.4. This is needed internally because of junit support, which was introduced in doctest 2.4.0.


Schneegans (approved):
Looks good!

Issue #62, opened 2019-09-29
Remove shader uniform location queries in the frame loop


JonasGilg:
A lot of shaders are querying the location of uniforms every frame, which should be avoided.
The uniform locations should be queries once after the shader has been linked and saved for reuse.

I propose the following convention for caching uniforms in a render class.

```c++
class SomeRenderer : public IVistaOpenGLDraw {
  public:
    SomeRenderer() {
        /* compile shader and link shader */
        mUniforms.mvp = glGetUniformLocation(mShader, "uMVP");
        mUniforms.color = glGetUniformLocation(mShader, "uColor");
        mUniforms.sunLocation = glGetUniformLocation(mShader, "uSunLocation");
    }

    /* ... */
  private:
    /* ... */

    uint32_t mShader;

    struct {
        uint32_t mvp;
        uint32_t color;
        uint32_t sunLocation;
    } mUniforms;
}
```

### cs-core
- [ ] tools/Mark.cpp

### cs-graphics
- [ ] MouseRay.cpp

### cs-gui
- [ ] ScreenSpaceGuiArea.cpp
- [ ] WorldSpaceGuiArea.cpp
- [ ] SpriteGui.cpp

### cs-scene
- [ ] Trajectory.cpp

### csp-atmospheres
- [ ] VistaAtmosphere.cpp

### csp-lod-bodies
- [ ] TileRenderer.cpp
- [ ] PlanetShader.cpp

### csp-measurement-tools
- [ ] PolygonTool.cpp
- [ ] DipStrikeTool.cpp
- [ ] EllipseTool.cpp
- [ ] PathTool.cpp

### csp-rings
- [ ] Ring.cpp

### csp-sharad
- [ ] Sharad.cpp

### csp-simple-bodies
- [ ] SimpleBody.cpp

### csp-stars
- [ ] Stars.cpp

### csp-trajectories
- [ ] DeepSpaceDot.cpp
- [ ] SunFlare.cpp

Schneegans:
This is definitely required. Usually, this should be in the scope of the OpenGL wrapper, but until we have a proper replacement for ViSTA, we could use the proposed approach.

TheJhonny007:
I will try and do it in conjunction with issue #27, since these issues are closely related.

TheJhonny007:
> This is definitely required. Usually, this should be in the scope of the OpenGL wrapper, but until we have a proper replacement for ViSTA, we could use the proposed approach.

[...]

PR #227, opened 2021-01-08
Remove uniform queries from the render loop

JonasGilg:
This fixes issue #62.


Schneegans (approved):
Looks good to me! Thank you for this heap of work!

(While testing this, I discovered a new and totally unrelated bug... stay tuned for an incoming review request :smile: )

PR #230, opened 2021-01-13
Frame Timing Plugin

Schneegans:
This PR adds a new plugin: `csp-timings`. I moved the existing frame statistics GUI to this plugin and added the functionality to export time series to a CSV file. This allows us to easily create graphs like this:

image


plugins/csp-timings/src/Plugin.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+void Plugin::update() {
+
+  // Enable or disable the statistics GUI item if necessary.
+  mGuiItem->setIsEnabled(mEnableStatistics && mFrameTimings->pEnableMeasurements.get());
+
+  // If frame timings are enabled, we may have to record them or update the on-screen statistics.
+  if (mFrameTimings->pEnableMeasurements.get()) {
+
+    // Get the frame timing information if either recording or the onscreen-statistics are enabled.
```
JonasGilg, line 114:
```suggestion
    // Get the frame timing information if either recording or the on-screen statistics are enabled.
```

plugins/csp-timings/src/Plugin.hpp
```diff
@@ -0,0 +1,40 @@
...
+  void deInit() override;
+  void update() override;
+
+ private:
+  /// This store the statistics GUI element.
+  std::unique_ptr<cs::gui::GuiItem> mGuiItem;
+
+  bool mEnableRecording  = false;
+  bool mEnableStatistics = false;
+  std::list<std::unordered_map<std::string, cs::utils::FrameTimings::QueryResult>> mRecordedTimings;
```
JonasGilg, line 33: What is the reasoning behind using a `std::list`? As it looks to me, a `std::vector` would be better, since we append only once a frame, but loop through twice a frame.

src/cs-core/GuiManager.cpp
```diff
@@ -81,13 +80,13 @@ GuiManager::GuiManager(std::shared_ptr<Settings> settings,
...
     // Create the global GUI area.
-    mGlobalGuiArea = std::make_unique<gui::WorldSpaceGuiArea>(
+    mGlobalGuiArea = std::make_shared<gui::WorldSpaceGuiArea>(
         mSettings->mGuiPosition->mWidthPixel, mSettings->mGuiPosition->mHeightPixel);
     mGlobalGuiArea->setUseLinearDepthBuffer(true);
   }
 
   // Create the local GUI area.
-  mLocalGuiArea = std::make_unique<gui::ScreenSpaceGuiArea>(pViewport);
+  mLocalGuiArea = std::make_shared<gui::ScreenSpaceGuiArea>(pViewport);
```
JonasGilg, line 89: What is the reasoning behind changing from `unique_ptr` to `shared_ptr`? The[ C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-unique) recommend using `unique_ptr` over `shared_ptr`.

plugins/csp-timings/src/Plugin.cpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    // Store the frame timing if we are in recording-mode.
+    if (mEnableRecording) {
+      mRecordedTimings.emplace_back(timings);
+    }
+  }
+
+  // Recording seems to have stopped last frame, so write the output file!
+  if (!mEnableRecording && mRecordedTimings.size() > 0) {
```
JonasGilg, line 150:
```suggestion
  if (!mEnableRecording && !mRecordedTimings.empty()) {
```

JonasGilg (changes requested):
Great work!

I think you should replace the record checkbox with a toggle button, this would be more intuitive. The plugin should also be enabled by default.

plugins/csp-measurement-tools/src/voronoi/BreakpointTree.cpp
```diff
@@ -123,13 +123,11 @@ Breakpoint* BreakpointTree::getNearestNode(double x, Breakpoint* current) const
...
                                                           : getNearestNode(x, current->mRightChild);
-  Breakpoint* nearest = current;
+  Breakpoint* nearest      = current;
 
   if (nearestChild &&
       (std::fabs(x - nearestChild->position().mX) < std::fabs(x - nearest->position().mX))) {
-    {
-      nearest = nearestChild;
-    }
+    { nearest = nearestChild; }
```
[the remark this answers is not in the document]
Schneegans: I don't know - seems like an old relict.

plugins/csp-timings/src/Plugin.hpp
Schneegans: We only `emplace_back()` once a frame during recording. This should be as fast as possible in order to reduce the impact on the timings. Then (after the recording) we loop through the list to retrieve the data. So I think `std::list` is a reasonable choice!

plugins/csp-timings/README.md
```diff
@@ -0,0 +1,26 @@
...
+    ...
+  }
+}
+```
+
+## Usage
+
+Once the plugin is loaded, you can enable the timer queries in the sidebar tab "Frame Timing".
+When the timer queries are enabled, you can either show the on-screen statistics or enable recording to a CSV file.
+When you disable the recording mode, a file called `timing-<current date>.csv` will be written to CosmoScout VR's `bin` directory.
```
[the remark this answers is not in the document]
Schneegans: I tried to make the wording more concise. The mentioned button will be implemented soon :smile:

src/cs-core/GuiManager.cpp
Schneegans: The these members can now be accessed through the interface of the `GuiManager`. Therefore I changed this.

As the `GuiManager` actually should own these, this is semantically not ideal. What should be done in this case? [Stackoverflow](https://stackoverflow.com/questions/43400850/return-a-unique-ptr-by-reference) suggests using raw pointers. Returning a reference to the `unique_ptr` seems weird, maybe we could simply return a `gui::ScreenSpaceGuiArea const&`?

JonasGilg: We can use references.

Otherwise we can use an `observer_ptr` (https://github.com/martinmoene/observer-ptr-lite), which is basically a raw pointer disguised as a smart pointer. As far as I know it might be standardized some time in the future.
https://en.cppreference.com/w/cpp/experimental/observer_ptr

JonasGilg (changes requested):
I am still not happy with the record button. It should somehow show, if it is recording. Maybe turn the dot red?

Schneegans:
The button uses the same styling as the radio buttons from the measurement tools. I now change it's text and icon to reflect the new state. What do you think?

JonasGilg (approved):
Looks nice!

plugins/csp-timings/src/Plugin.hpp
JonasGilg: I know we already merged, but my mind couldn't let go 🙈. `std::list` should still perform worse, since it has to allocate memory every time you insert a node, while `std::vector` only needs to allocate memory if the underlying array is full. Memory allocations are very very expensive. I did a small benchmark, here are the results:

std::list: 
```c++
void list() {
    spdlog::info("list");
    spdlog::stopwatch sw;
    std::list<std::array<uint64_t, 10>> l;

    for (size_t i = 0; i < ELEMENTS; ++i) {
        std::array<uint64_t, 10> a{};

        for (auto&& e : a) {
            e = i;
        }

        l.emplace_back(a);
    }

    spdlog::info("  insert: {}", sw);

    uint64_t sum = 0;
    for (auto const& e : l) {
        sum += std::accumulate(e.begin(), e.end(), 0ull);
    }

    spdlog::debug("    Sum: {}", sum);
    spdlog::info("complete: {}\n", sw);
}
```

std::vector:
```c++
void vector() {
    spdlog::info("vector");
    spdlog::stopwatch sw;
    std::vector<std::array<uint64_t, 10>> l;

    for (size_t i = 0; i < ELEMENTS; ++i) {
        std::array<uint64_t, 10> a{};

        for (auto&& e : a) {
            e = i;
        }

        l.emplace_back(a);
    }

    spdlog::info("  insert: {}", sw);

    uint64_t sum = 0;
    for (auto const& e : l) {
        sum += std::accumulate(e.begin(), e.end(), 0ull);
    }

    spdlog::debug("    Sum: {}", sum);
    spdlog::info("complete: {}\n", sw);
}
```

Results with 10000000 elements:
```
list
  insert: 5.9572255
complete: 8.2400265

vector
  insert: 3.3486662
complete: 5.0138386
```

I think `std::vector` should be preferred here.

Schneegans: While I agree that this is an interesting question and that you are maybe right - I hope that you agree that this is definitely a [case of premature optimization](https://stackify.com/premature-optimization-evil/)! Therefore I won't spent much time here...

What happens if you replace the `emplace_back(a)` with a `emplace_back(std::move(a))` (or `push_back(std::move(a))`, this shouldn't make any difference)? `std::arrays` are completely stack-allocated, aren't they? So moving them shouldn't make any difference - you could try using `std::unordered_map` as we are using in the actual code they should be faster to move as most of the memory is heap-allocated (I think).

JonasGilg: I agree, that we shouldn't waste time over a single line like this, but I think it might be worth to get a better understanding in the long term. The [Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rsl-vector) say that `std::vector` should nearly always be preferred. So I would say it is a premature optimization to use `std::list` at all.

`std::vector<uint64_t, 10>` has actually the same size as `std::unordered_map`, which is 80 bytes.

I think we can leave it as is, since the performance impact here is close to none.

PR #232, opened 2021-01-21
Fix energy conservation in shading

Schneegans:
In HDR rendering, we are missing a division by π to be ernergy-conserving for Lambertian surfaces. See https://seblagarde.wordpress.com/2012/01/08/pi-or-not-to-pi-in-game-lighting-equation/ for an explanation.

The brightness of the clouds in HDR mode is still quite arbitrary. This needs to be fixed in the future.

### DIFF
--- a/plugins/csp-rings/src/Ring.cpp
+++ b/plugins/csp-rings/src/Ring.cpp
@@ -122,7 +122,7 @@ Ring::Ring(std::shared_ptr<cs::core::Settings> settings,
   mSphereVAO.SpecifyAttributeArrayFloat(
       0, 2, GL_FLOAT, GL_FALSE, sizeof(glm::vec2), 0, &mSphereVBO);
 
-  // Recreate the shader if HDR rendering mode are toggled.
+  // Recreate the shader if HDR rendering mode is toggled.
   mEnableHDRConnection =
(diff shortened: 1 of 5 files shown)


plugins/csp-rings/src/Ring.cpp
```diff
@@ -112,17 +122,9 @@ Ring::Ring(std::shared_ptr<cs::core::Settings> settings,
...
-  mShader.InitFragmentShaderFromString(SPHERE_FRAG);
-  mShader.Link();
-
-  mUniforms.modelViewMatrix  = mShader.GetUniformLocation("uMatModelView");
-  mUniforms.projectionMatrix = mShader.GetUniformLocation("uMatProjection");
-  mUniforms.surfaceTexture   = mShader.GetUniformLocation("uSurfaceTexture");
-  mUniforms.radii            = mShader.GetUniformLocation("uRadii");
-  mUniforms.farClip          = mShader.GetUniformLocation("uFarClip");
-  mUniforms.sunIlluminance   = mShader.GetUniformLocation("uSunIlluminance");
+  // Recreate the shader if HDR rendering mode are toggled.
```
JonasGilg, line 125:
```suggestion
  // Recreate the shader if HDR rendering mode is toggled.
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #237, opened 2021-02-08
Feature/wms overlays

TheFridge76:


Schneegans:
Thank you very much for this awesome addition! Could you please merge the current `develop`? You'll need to rebuild the externals. Also the CEF version changed, so maybe you have to delete `build\windows-externals-Release\cef`.

Schneegans:
You may also add this plugin to the list of plugins in the README.md :smile:

Schneegans:
I just tested this on Linux and all seems to work nicely. I'll look at the code hopefully later this week - here are some initial observations / suggestions:

* The loading of the plugin takes a while... can this be improved? Especially on first load at least some progress output to the console would be nice. Else it feels like the application hangs.
* I observed quite a few warnings that for a specific request the time dimension was invalid (especially at the beginning and the end of the valid time ranges). Can this be avoided?
* I think the `layers` icon would be more fitting than `panorama`.
* I'll think about the sidebar UI, maybe we can improve it a little. It looks a bit cluttered but I have no good idea for improvements yet :smile: .
* I think we should start using a unified cache structure. I would suggest `cache/<csp-plugin-name>/whatever`. So the LOD-Bodies would write to `cache/csp-lod-bodies/` and your plugin would write to `cache/csp-wms-overlays/capability-cache` and to `cache/csp-wms-overlays/texture-cache`. This way we only have one top-level cache directory! 

Thank you once more for this great addition!

TheFridge76:
Thanks for your feedback on the plugin!
To address some of your suggestions:

- When loading the plugin, the capabilities of all WMS servers given in the config file are queried. Apparently the connection to the server can take quite a lot of time while the actual download and parsing of the file is much quicker as far as i know. I don't know if we can query the progress of the connection, but I could print a message to the console before and after loading each server's capabilities. Perhaps it could also be worth it to look into parallelizing the requests.
- At least for the DWD server "invalid time dimension" warnings can occur near the beginning of the time range because for some layers old data seems to be discarded as new data becomes available. However, I think there might also be an issue with the prefetching mechanism, that causes requests for invalid timesteps. I'll look into that.
- The unified cache structure sounds like a good idea. Do you think we should remove the configuration entries for specifying the directories in that case, or should the new structure just be the default value? And should the cache still be located in bin/ or maybe rather be moved to share/?

Schneegans:
Have you checked the `Downloader` class? It already supports parallel downloads and some progress output. It's used here: https://github.com/cosmoscout/cosmoscout-vr/blob/develop/src/cosmoscout/Application.cpp#L330

Regarding the cache: I think we should just change the default configuration and moving it to `share/cache` seems reasonable!

plugins/csp-wms-overlays/gui/css/csp-wms-overlays.css
```diff
@@ -0,0 +1,32 @@
+#wmsOverlays\.legendContainer {
```
Schneegans, line 1: In my opinion using dots in container IDs is a bit unfortunate as it is easily confused with classes. I would suggest using dashes: `#wmsOverlays-legendContainer`. This requires several changes which I will not mark all in this review :smile:

plugins/csp-wms-overlays/src/Plugin.hpp
```diff
@@ -0,0 +1,135 @@
...
+      cs::utils::DefaultProperty<std::string> mActiveLayer{
+          "None"}; ///< The name of the currently active WMS layer.
+      cs::utils::DefaultProperty<std::string> mActiveStyle{
+          ""}; ///< The name of the style for the currently active WMS layer.
+      cs::utils::DefaultProperty<Bounds> mActiveBounds{
+          {-180., 180., -90., 90.}}; ///< The bounds for the currently active WMS layer.
+      std::vector<std::string> mWms; ///<	URLs of WMS servers.
+    };
+
+    std::map<std::string, Body> mBodies; ///< A list of bodies with their anchor names.
```
Schneegans, line 71: I would move the comments above the corresponding lines. This is more clang-format-friendly.

plugins/csp-wms-overlays/src/Plugin.hpp
```diff
@@ -0,0 +1,135 @@
...
+  std::map<std::string, std::optional<WebMapService>>
+      mActiveServers; ///< The currently active WebMapService for each center name.
+  std::map<std::string, std::optional<WebMapLayer>>
+      mActiveLayers; ///< The currently active WebMapLayer for each center name.
+
+  bool mNoMovement;                ///< True when the observer is not moving.
+  bool mNoMovementRequestedUpdate; ///< True when the active overlay was requested to update its
+                                   ///< bounds because the observer is not moving.
+  std::chrono::time_point<std::chrono::high_resolution_clock>
+      mNoMovementSince; ///< Time at which the observer stopped moving.
```
Schneegans, line 124: I would move the comments above the corresponding lines. This is more clang-format-friendly.

plugins/csp-wms-overlays/src/logger.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "logger.hpp"
+
+#include "../../../src/cs-utils/logger.hpp"
+
+namespace csp::wmsoverlays {
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+spdlog::logger& logger() {
+  static auto logger = cs::utils::createLogger("csp-wms-overlay");
```
Schneegans, line 16:
```suggestion
  static auto logger = cs::utils::createLogger("csp-wms-overlays");
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.hpp
```diff
@@ -0,0 +1,92 @@
+////////////////////////////////////////////////////////////////////////////////////////////////////
+//                               This file is part of CosmoScout VR                               //
+//      and may be used under the terms of the MIT license. See the LICENSE file for details.     //
+//                        Copyright: (c) 2019 German Aerospace Center (DLR)                       //
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef CSP_WMS_TEXTURE_LOADER_HPP
+#define CSP_WMS_TEXTURE_LOADER_HPP
```
Schneegans, line 8:
```suggestion
#ifndef CSP_WMS_OVERLAYS_TEXTURE_LOADER_HPP
#define CSP_WMS_OVERLAYS_TEXTURE_LOADER_HPP
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.hpp
```diff
@@ -0,0 +1,92 @@
...
+  const std::map<std::string, std::string> mMimeToExtension = {
+      {"image/png", "png"}, {"image/jpeg", "jpg"}};
+
+  std::mutex            mTextureMutex;
+  cs::utils::ThreadPool mThreadPool;
+};
+
+} // namespace csp::wmsoverlays
+
+#endif // CSP_WMS_TEXTURE_LOADER_HPP
```
Schneegans, line 92:
```suggestion
#endif // CSP_WMS_OVERLAYS_TEXTURE_LOADER_HPP
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.hpp
```diff
@@ -0,0 +1,92 @@
...
+ public:
+  /// Struct for defining parameters for a request to a WMS.
+  struct Request {
+    int                        mMaxSize;
+    std::string                mStyle;
+    Bounds                     mBounds;
+    std::optional<std::string> mTime;
+  };
+
+  /// Creates a new ThreadPool with the specified amount of threads.
```
Schneegans, line 40: Really? :smile:

plugins/csp-wms-overlays/src/WebMapTextureLoader.hpp
```diff
@@ -0,0 +1,92 @@
...
+  WebMapTextureLoader();
+
+  ~WebMapTextureLoader();
+
+  /// Async WMS texture loader.
+  /// Returns an empty optional if loading the texture failed.
+  std::future<std::optional<WebMapTexture>> loadTextureAsync(WebMapService const& wms,
+      WebMapLayer const& layer, Request const& request, std::string const& mapCache,
+      bool const& saveToCache);
```
Schneegans, line 49: No need to pass built-ins as reference.

```suggestion
      bool saveToCache);
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+  /// WMS texture loader.
+  /// Returns an empty optional if loading the texture failed.
+  std::optional<WebMapTexture> loadTexture(WebMapService const& wms, WebMapLayer const& layer,
+      Request const& request, std::string const& mapCache, bool const& saveToCache);
```
Schneegans, line 54:
```suggestion
      Request const& request, std::string const& mapCache, bool saveToCache);
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  inline bool operator==(const Bounds& rhs) const {
+    return mMinLon == rhs.mMinLon && mMaxLon == rhs.mMaxLon && mMinLat == rhs.mMinLat &&
+           mMaxLat == rhs.mMaxLat;
+  }
+};
+
+/// Struct for the duration of the WMS time step.
+///	Ideally only one of the members should be non-zero.
```
Schneegans, line 53:
```suggestion
/// Ideally only one of the members should be non-zero.
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+struct Duration {
+  int                              mYears        = 0;
+  int                              mMonths       = 0;
+  boost::posix_time::time_duration mTimeDuration = boost::posix_time::seconds(0);
+
+  bool isDuration() const;
```
Schneegans, line 59: Maybe a comment would be helpful here.

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+  Duration                 mSampleDuration; ///< The duration of one sample in WMS interval.
+
+  inline bool operator==(const TimeInterval& rhs) const {
+    return mStartTime == rhs.mStartTime && mEndTime == rhs.mEndTime && mFormat == rhs.mFormat &&
+           mSampleDuration == rhs.mSampleDuration;
+  }
+};
+
+namespace utils {
+
```
Schneegans, line 80: I would really appreciate a high-level comment at this point. Which time formats are used below? Why is this different than the stuff in cs::utils::convert::time?

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+/// Adds the interval multiple times, if it is specified (e.g. for pre-fetch).
+boost::posix_time::ptime addDurationToTime(
+    boost::posix_time::ptime time, Duration duration, int multiplier = 1);
+
+/// Tries to get the value contained in a XML element.
+/// Starts at baseElement and then descends into the children given as childPath.
+/// The return value is empty if the requested element is not present.
+template <typename T>
+std::optional<T> getElementValue(
+    VistaXML::TiXmlElement* baseElement, std::vector<std::string> childPath = {}) {
```
Schneegans, line 113: Can this be a reference?
```suggestion
    VistaXML::TiXmlElement* baseElement, std::vector<std::string> const& childPath = {}) {
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+  return {};
+}
+
+/// Tries to get the value of an integer attribute representing a size.
+/// The returned value (inner optional) is empty if the attribute specifies an unlimited size.
+/// The return value (outer optional) is empty if the requested attribute is not present.
+std::optional<std::optional<int>> getSizeAttribute(
+    VistaXML::TiXmlElement* element, std::string attributeName);
```
Schneegans, line 138: Also here:
```suggestion
    VistaXML::TiXmlElement* element, std::string const& attributeName);
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+/// Gets the value of the given attribute on the given element.
+/// The return value is empty if the requested element is not present.
+template <typename T>
+std::optional<T> getAttribute(VistaXML::TiXmlElement* element, std::string attributeName) {
```
Schneegans, line 143:
```suggestion
std::optional<T> getAttribute(VistaXML::TiXmlElement* element, std::string const& attributeName) {
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+  if (result == VistaXML::TIXML_SUCCESS) {
+    return value;
+  }
+  return {};
+}
+
+/// QueryValueAttribute does not work for strings containing spaces, so Attribute is used instead.
+template <>
+std::optional<std::string> getAttribute<std::string>(
+    VistaXML::TiXmlElement* element, std::string attributeName);
```
Schneegans, line 155:
```suggestion
    VistaXML::TiXmlElement* element, std::string const& attributeName);
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+/// Booleans may be given as either integers (0->false, 1->true) or strings.
+template <>
+std::optional<bool> getAttribute<bool>(VistaXML::TiXmlElement* element, std::string attributeName);
```
Schneegans, line 159:
```suggestion
std::optional<bool> getAttribute<bool>(VistaXML::TiXmlElement* element, std::string const& attributeName);
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+/// Sets the given var to the value of the optional, if it is present.
+template <typename T>
+void setOrKeep(T& var, std::optional<T> optional) {
```
Schneegans, line 163:
```suggestion
void setOrKeep(T& var, std::optional<T> const& optional) {
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+  var = optional.value_or(var);
+}
+
+/// Sets the given var to the value of the optional, if it is present.
+template <typename T>
+void setOrKeep(std::optional<T>& var, std::optional<T> optional) {
```
Schneegans, line 169:
```suggestion
void setOrKeep(std::optional<T>& var, std::optional<T> const& optional) {
```

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+#include <curlpp/Easy.hpp>
+#include <curlpp/Options.hpp>
+
+namespace csp::wmsoverlays {
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+WebMapService::WebMapService(std::string url, std::string cacheDir)
+    : mUrl(url)
+    , mCacheDir(cacheDir)
```
Schneegans, line 27:
```suggestion
    : mUrl(std::move(url))
    , mCacheDir(std::move(cacheDir))
```

plugins/csp-wms-overlays/src/WebMapService.hpp
```diff
@@ -0,0 +1,88 @@
...
+  /// Gets the general settings of the service.
+  Settings getSettings() const;
+
+  /// Gets the root layer of the service.
+  WebMapLayer getRootLayer() const;
+  /// Gets a list of all layers of the service, for which maps can be requested.
+  std::vector<WebMapLayer> getLayers() const;
+  /// Gets the layer with the given name, if one exists.
+  /// Returns an empty optional otherwise.
+  std::optional<WebMapLayer> getLayer(std::string name) const;
```
Schneegans, line 51: These should all return by `const&`, shouldn't they? And the last should take a `const&` string.

```suggestion
  std::string const& getUrl() const;
  /// Gets a brief description of the service.
  std::string const& getTitle() const;

  /// Gets the general settings of the service.
  Settings const& getSettings() const;

  /// Gets the root layer of the service.
  WebMapLayer const& getRootLayer() const;
  /// Gets a list of all layers of the service, for which maps can be requested.
  std::vector<WebMapLayer> const& getLayers() const;
  /// Gets the layer with the given name, if one exists.
  /// Returns an empty optional otherwise.
  std::optional<WebMapLayer> const& getLayer(std::string const& name) const;
```

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+std::vector<WebMapLayer> WebMapService::getLayers() const {
+  return mRequestableLayers;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::optional<WebMapLayer> WebMapService::getLayer(std::string name) const {
+  std::vector<WebMapLayer> layers = getLayers();
+  auto                     layer  = std::find_if(
+      layers.begin(), layers.end(), [&name](WebMapLayer l) { return l.getName() == name; });
```
Schneegans, line 71: Once `name` is passed as `const&` you can omit the `&` here:
```suggestion
      layers.begin(), layers.end(), [name](WebMapLayer l) { return l.getName() == name; });
```

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+  if (layer == layers.end()) {
+    return {};
+  } else {
+    return *layer;
+  }
```
Schneegans, line 76:
```suggestion
  }

  return *layer;
```

plugins/csp-wms-overlays/src/WebMapLayer.hpp
```diff
@@ -0,0 +1,85 @@
...
+    std::optional<int>         mFixedHeight; ///< Only textures with this height may be requested.
+    Bounds                     mBounds;      ///< Default (maximum) bounds of the layer.
+    bool                       mOpaque = false; ///< Specifies whether the layer is opaque.
+    std::vector<Style>         mStyles;         ///< List of styles for the layer.
+    std::vector<TimeInterval>  mTimeIntervals;  ///< TimeIntervals, for which data is available.
+    std::optional<std::string> mAttribution;    ///< Attribution for the layer.
+    std::optional<double>
+        mMinScale; ///< Minimum scale denominator for which it is appropriate to generate a map.
+    std::optional<double>
+        mMaxScale; ///< Maximum scale denominator for which it is appropriate to generate a map.
```
Schneegans, line 52: I would move the comments above the lines.

plugins/csp-wms-overlays/src/WebMapLayer.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include <curlpp/Easy.hpp>
+#include <curlpp/Options.hpp>
+
+namespace csp::wmsoverlays {
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+WebMapLayer::WebMapLayer(VistaXML::TiXmlElement* element, Settings settings)
+    : mSettings(settings) {
```
Schneegans, line 21: This should either be moved (if possible, I am not sure) or passed per `const&`.

```suggestion
    : mSettings(std::move(settings)) {
```

plugins/csp-wms-overlays/src/WebMapLayer.hpp
```diff
@@ -0,0 +1,85 @@
...
+  std::string getName() const;
+  /// Gets a narrative description of the layer.
+  std::optional<std::string> getAbstract() const;
+  /// Gets the general settings of the layer.
+  Settings getSettings() const;
+
+  /// Checks if map data may be requested for the layer.
+  bool isRequestable() const;
+  /// Gets a list of all child layers of the layer.
+  std::vector<WebMapLayer> getAllLayers() const;
```
Schneegans, line 69: These should all return by `const&` if they do not return temporaries.

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,89 @@
...
+  WebMapException(VistaXML::TiXmlElement* element);
+
+  inline bool operator!=(const WebMapException& rhs) const {
+    return mCode != rhs.mCode || mText != rhs.mText;
+  }
+
+  /// Get the code identifying the type of exception that occurred.
+  Code getCode() const;
+  /// Get a short description of the error.
+  std::string getText() const;
```
Schneegans, line 55:
```suggestion
  std::string const& getText() const;
```

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,89 @@
...
+/// Class to store a collection of WMS exceptions.
+class WebMapExceptionReport : public std::exception {
+ public:
+  /// Construct a WebMapExceptionReport from a XML document.
+  WebMapExceptionReport(VistaXML::TiXmlDocument doc);
+  /// Construct a WebMapExceptionReport from a string containing a XML document.
+  WebMapExceptionReport(std::string const& xml);
+
+  /// Gets a list of WMS exceptions that occurred.
+  std::vector<WebMapException> getExceptions() const;
```
Schneegans, line 75:
```suggestion
  std::vector<WebMapException> const& getExceptions() const;
```

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,89 @@
...
+                            ///< request is equal to current value of service metadata update
+                            ///< sequence number.
+    eInvalidUpdateSequence, ///< Value of (optional) UpdateSequence parameter in GetCapabilities
+                            ///< request is greater than current value of service metadata update
+                            ///< sequence number.
+    eMissingDimensionValue, ///< Request does not include a sample dimension value, and the server
+                            ///< did not declare a default value for that dimension.
+    eInvalidDimensionValue, ///< Request contains an invalid sample dimension value.
+    eOperationNotSupported, ///< Request is for an optional operation that is not supported by the
+                            ///< server.
```
Schneegans, line 43: I think moving the comments above could look better.

plugins/csp-wms-overlays/src/ShadersTextureOverlay.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // ===========================================================================
+    void main()
+    {     
+        float fDepth = GetDepth();
+        if (fDepth == 1000000.0) 
+        {
+            discard;
+        }else{
+            dvec3 worldPos  = GetPosition();
+            //dvec2 lnglat    = GetLngLat(worldPos);
```
Schneegans, line 193:
```suggestion
            dvec3 worldPos = GetPosition();
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,141 @@
+#ifndef TEXTURE_OVERLAY_RENDERER
+#define TEXTURE_OVERLAY_RENDERER
```
Schneegans, line 2:
```suggestion
////////////////////////////////////////////////////////////////////////////////////////////////////
//                               This file is part of CosmoScout VR                               //
//      and may be used under the terms of the MIT license. See the LICENSE file for details.     //
//                        Copyright: (c) 2020 German Aerospace Center (DLR)                       //
////////////////////////////////////////////////////////////////////////////////////////////////////

#ifndef CSP_WMS_OVERLAYS_TEXTURE_OVERLAY_RENDERER_HPP
#define CSP_WMS_OVERLAYS_TEXTURE_OVERLAY_RENDERER_HPP
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,141 @@
...
+  std::array<float, 3> mMaxBounds; ///< Upper Corner of the bounding volume for the planet.
+
+  bool mShaderDirty        = true;
+  int  mLightingConnection = -1;
+  int  mHDRConnection      = -1;
+};
+
+} // namespace csp::wmsoverlays
+
+#endif // TEXTURE_OVERLAY_RENDERER
```
Schneegans, line 141:
```suggestion
#endif // CSP_WMS_OVERLAYS_TEXTURE_OVERLAY_RENDERER_HPP
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,690 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+TextureOverlayRenderer::TextureOverlayRenderer(std::string center,
+    std::shared_ptr<cs::core::SolarSystem>                 solarSystem,
+    std::shared_ptr<cs::core::TimeControl>                 timeControl,
+    std::shared_ptr<cs::core::Settings> settings, std::shared_ptr<Plugin::Settings> pluginSettings)
+    : mSettings(settings)
+    , mPluginSettings(pluginSettings)
+    , mCenterName(center)
```
Schneegans, line 49:
```suggestion
    : mSettings(std::move(settings))
    , mPluginSettings(std::move(pluginSettings))
    , mCenterName(std::move(center))
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,690 @@
...
+    , mWMSTexture(new VistaTexture(GL_TEXTURE_2D))
+    , mSecondWMSTexture(new VistaTexture(GL_TEXTURE_2D))
+    , mSolarSystem(solarSystem)
+    , mTimeControl(timeControl)
```
Schneegans, line 53:
```suggestion
    , mSolarSystem(std::move(solarSystem))
    , mTimeControl(std::move(timeControl))
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,141 @@
...
+class TextureOverlayRenderer : public IVistaOpenGLDraw {
+ public:
+  TextureOverlayRenderer(std::string center, std::shared_ptr<cs::core::SolarSystem> solarSystem,
+      std::shared_ptr<cs::core::TimeControl> timeControl,
+      std::shared_ptr<cs::core::Settings>    settings,
+      std::shared_ptr<Plugin::Settings>      pluginSettings);
+  virtual ~TextureOverlayRenderer();
+
+  /// Returns the SPICE name of the body to which this renderer is assigned.
+  std::string getCenter() const;
```
Schneegans, line 45:
```suggestion
  std::string const& getCenter() const;
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+  /// Configures the internal renderer according to the given values.
+  void configure(Plugin::Settings::Body const& settings);
```
Schneegans, line 48: If `Plugin::Settings::Body` can be moved, I wouldn't use `const&` as this method copies the value. So the interface would be clearer without the ref.
```suggestion
  void configure(Plugin::Settings::Body settings);
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,690 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::string TextureOverlayRenderer::getCenter() const {
+  return mCenterName;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+void TextureOverlayRenderer::configure(Plugin::Settings::Body const& settings) {
+  mSimpleWMSOverlaySettings = settings;
```
Schneegans, line 136:
```suggestion
void TextureOverlayRenderer::configure(Plugin::Settings::Body settings) {
  mSimpleWMSOverlaySettings = std::move(settings);
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,690 @@
...
+
+  clearTextures();
+
+  VistaSceneGraph* pSG = GetVistaSystem()->GetGraphicsManager()->GetSceneGraph();
+  pSG->GetRoot()->DisconnectChild(mGLNode.get());
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::string TextureOverlayRenderer::getCenter() const {
```
Schneegans, line 129:
```suggestion
std::string const& TextureOverlayRenderer::getCenter() const {
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,690 @@
...
+
+  mActiveWMS.reset();
+  mActiveWMSLayer.reset();
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+void TextureOverlayRenderer::setStyle(std::string style) {
+  if (mStyle != style) {
+    mStyle = style;
```
Schneegans, line 176:
```suggestion
    mStyle = std::move(style);
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+  std::unique_ptr<VistaOpenGLNode> mGLNode;
+
+  VistaGLSLShader mShader; ///< Vista GLSL shader object used for rendering
+
+  static const std::string SURFACE_GEOM; ///< Code for the geometry shader
+  static const std::string SURFACE_VERT; ///< Code for the vertex shader
+  static const std::string SURFACE_FRAG; ///< Code for the fragment shader
+
+  std::unordered_map<VistaViewport*, VistaTexture*>
```
Schneegans, line 101: I would move most of the comments here above the corresponding lines.

Schneegans (changes requested):
Fuah - this was _alot_. But apart from several `const&` it looks great as far as I can tell. I'll do some more testing on my end. Could you also run `clang-tidy.sh` in a separate commit? I guess it will find alot of stuff. But please commit only the resulting changes in `plugins/csp-wms-overlays` :smile:

Schneegans:
It also appears to me that the capabilities cache is outdated quite frequently. If you download the capabilities in parallel we could maybe trying not to cache this? Most awesome solution would be to parse the capabilities document directly after downloading directly in the download thread. Only once this is finished, the UI should be filled with the newly available data sets. This way the plugin loading would be much faster!

TheFridge76:
The capability cache does bring some load time improvements at least when frequently restarting CosmoScout, so I don't think we should stop caching the capabilities. The NASA NEO server unfortunately does not specify a sequence number, so its capabilites have to be downloaded each time the plugin is loaded, but both the NASA SVS server and the DWD server profited from using the cache in a short test I did:

|Server|without Cache|with Cache|
|:-|-:|-:|
|NEO|1.7-2.9s|-|
|SVS|4.4-5s|1,7-2s|
|DWD|5.9-6s|0.5-0.7s|

I'm sorry, but I didn't quite get the second part of your comment: Do you suggest to proceed with the initialization of CosmoScout while the capabilities are downloaded and parsed in the background? If we added more work to be done to the download threads, this would also mean that the `Downloader` class can't be used anymore, right?

Schneegans:
I just tested this with the DWD server. The update sequence number is not changed when the time ranges change. It still reports `13431` as it did this morning. However, the valid time dimensions for various layers have changed in the mean time. So for example, the "Niederschlag" layer only shows the most recent data if I delete the cache by hand.

And yes, you're right using the `Downloader` wouldn't be straight-forward anymore. You would either have to copy the relevant  code from this class (it's not much. actually just a combination of the `Threadpool` and `cs::utils::filesystem::downloadFile()`) or you launch a new parsing thread once the download finished...

What do you think?

TheFridge76:
Ah, right! I guess with DWD not updating the updateSequence often enough the caching would only be useful for the SVS server. In this case I think it makes sense to remove the capability caching for now.

I'll work on implementing the parallel download in the next days. I think using `Threadpool` it should be quite straight-forward. However, as I'm downloading the file directly to in-memory streams I will keep using the manual curlpp request setup instead of `cs::utils::filesystem::downloadFile()` (Maybe a method for downloading files to any `std::ostream` should be added to `cs::utils` at some point?)

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,692 @@
...
+  double farClip  = NAN;
+  GetVistaSystem()
+      ->GetDisplayManager()
+      ->GetCurrentRenderInfo()
+      ->m_pViewport->GetProjection()
+      ->GetProjectionProperties()
+      ->GetClippingRange(nearClip, farClip);
+
+  // copy depth buffer from previous rendering
+  GLint iViewport[4];
```
JonasGilg, line 553: Use `std::array`

plugins/csp-wms-overlays/src/TextureOverlayRenderer.cpp
```diff
@@ -0,0 +1,692 @@
...
+  auto* viewport = GetVistaSystem()->GetDisplayManager()->GetCurrentRenderInfo()->m_pViewport;
+  VistaTexture* depthBuffer = mDepthBufferData[viewport];
+
+  depthBuffer->Bind();
+  glCopyTexImage2D(GL_TEXTURE_RECTANGLE, 0, GL_DEPTH_COMPONENT, iViewport[0], iViewport[1],
+      iViewport[2], iViewport[3], 0);
+
+  // get matrices and related values
+  GLfloat glMatP[16];
+  GLfloat glMatMV[16];
```
JonasGilg, line 565: Use `std::array`

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,167 @@
...
+/// Therefore it copies the depth buffer first. Second, in the shader it does an inverse projection
+/// to get the cartesian coordinates. This coordinates are transformed to latitude and longitude to
+/// do the lookup in the geo-referenced texture. The value is then overlayed on that pixel position.
+class TextureOverlayRenderer : public IVistaOpenGLDraw {
+ public:
+  TextureOverlayRenderer(std::string center, std::shared_ptr<cs::core::SolarSystem> solarSystem,
+      std::shared_ptr<cs::core::TimeControl> timeControl,
+      std::shared_ptr<cs::core::Settings>    settings,
+      std::shared_ptr<Plugin::Settings>      pluginSettings);
+  virtual ~TextureOverlayRenderer();
```
JonasGilg, line 48: Use either override or final for inherited virtual functions.
```suggestion
  ~TextureOverlayRenderer() override;
```
```suggestion
  ~TextureOverlayRenderer() final;
```

plugins/csp-wms-overlays/src/TextureOverlayRenderer.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /// The bounds will be updated the next time the Do() method is called.
+  void requestUpdateBounds();
+
+  /// The current map bounds of this overlay.
+  /// This may be used for setting the bounds.
+  cs::utils::Property<Bounds> pBounds;
+
+  /// Interface implementation of IVistaOpenGLDraw
+  virtual bool Do();
+  virtual bool GetBoundingBox(VistaBoundingBox& bb);
```
JonasGilg, line 75: Use `override` or `final` for inherited virtual functions.

plugins/csp-wms-overlays/src/WebMapException.cpp
```diff
@@ -0,0 +1,169 @@
...
+      code = WebMapException::Code::eOperationNotSupported;
+    }
+  }
+  return in;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::ostream& operator<<(std::ostream& out, WebMapException::Code& code) {
+  std::string codeStr = "UnknownCode";
```
JonasGilg, line 49: This is used nowhere.

```suggestion
```

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,97 @@
...
+    /// Request does not include a sample dimension value, and the server did not declare a default
+    /// value for that dimension.
+    eMissingDimensionValue,
+    /// Request contains an invalid sample dimension value.
+    eInvalidDimensionValue,
+    /// Request is for an optional operation that is not supported by the server.
+    eOperationNotSupported,
+  };
+
+  WebMapException(VistaXML::TiXmlElement* element);
```
JonasGilg, line 54:
```suggestion
  explicit WebMapException(VistaXML::TiXmlElement* element);
```

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,97 @@
...
+    return mCode != rhs.mCode || mText != rhs.mText;
+  }
+
+  /// Get the code identifying the type of exception that occurred.
+  Code getCode() const;
+  /// Get a short description of the error.
+  std::string const& getText() const;
+
+  /// Get type and description of the error as one string.
+  virtual const char* what() const noexcept;
```
JonasGilg, line 66: Use `override` or `final` instead of `virtual`.

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,97 @@
...
+  Code        mCode;
+  std::string mText;
+  std::string mMessage;
+};
+
+/// Class to store a collection of WMS exceptions.
+class WebMapExceptionReport : public std::exception {
+ public:
+  /// Construct a WebMapExceptionReport from a XML document.
+  WebMapExceptionReport(VistaXML::TiXmlDocument doc);
```
JonasGilg, line 78:
```suggestion
  explicit WebMapExceptionReport(VistaXML::TiXmlDocument doc);
```

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,97 @@
...
+  /// Construct a WebMapExceptionReport from a string containing a XML document.
+  WebMapExceptionReport(std::string const& xml);
```
JonasGilg, line 80:
```suggestion
  explicit WebMapExceptionReport(std::string const& xml);
```

plugins/csp-wms-overlays/src/WebMapException.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+  /// Gets a list of WMS exceptions that occurred.
+  std::vector<WebMapException> const& getExceptions() const;
+
+  /// Gets a single string describing all WMS exceptions that occurred.
+  virtual const char* what() const noexcept;
```
JonasGilg, line 86: Use `override` or `final` instead of `virtual`.

plugins/csp-wms-overlays/src/WebMapLayer.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include <curlpp/Easy.hpp>
+#include <curlpp/Options.hpp>
+
+namespace csp::wmsoverlays {
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+WebMapLayer::WebMapLayer(VistaXML::TiXmlElement* element, Settings settings)
+    : mSettings(std::move(settings)) {
+  VistaXML::TiXmlHandle elementHandle(element);
```
JonasGilg, line 22: This is never used.

```suggestion
```

plugins/csp-wms-overlays/src/WebMapLayer.cpp
```diff
@@ -0,0 +1,185 @@
...
+       crsElement                         = crsElement->NextSiblingElement("CRS")) {
+    std::optional<std::string> crs = utils::getElementValue<std::string>(crsElement);
+    if (crs.has_value() && !cs::utils::contains(mSettings.mCrs, crs.value())) {
+      mSettings.mCrs.push_back(crs.value());
+    }
+  }
+
+  for (VistaXML::TiXmlElement* layerElement = element->FirstChildElement("Layer"); layerElement;
+       layerElement                         = layerElement->NextSiblingElement("Layer")) {
+    mSubLayers.push_back(WebMapLayer(layerElement, mSettings));
```
JonasGilg, line 110:
```suggestion
    mSubLayers.emplace_back(layerElement, mSettings);
```

plugins/csp-wms-overlays/src/WebMapLayer.cpp
```diff
@@ -0,0 +1,185 @@
...
+  return mSubLayers;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+void WebMapLayer::getRequestableLayers(std::vector<WebMapLayer>& layers) const {
+  if (isRequestable()) {
+    layers.push_back(*this);
+  }
+  for (WebMapLayer sub : mSubLayers) {
```
JonasGilg, line 158:
```suggestion
  for (WebMapLayer const& sub : mSubLayers) {
```

plugins/csp-wms-overlays/src/WebMapLayer.hpp
```diff
@@ -0,0 +1,95 @@
...
+  /// Struct for storing information on a single style for a WMS layer.
+  struct Style {
+    /// Internal name of the style for requests.
+    const std::string mName;
+    /// Human readable description of the style.
+    const std::string mTitle;
+    /// URL at which a legend image may be found.
+    const std::optional<std::string> mLegendUrl;
+
+    Style(VistaXML::TiXmlElement* element);
```
JonasGilg, line 33:
```suggestion
    explicit Style(VistaXML::TiXmlElement* element);
```

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+std::vector<WebMapLayer> const& WebMapService::getLayers() const {
+  return mRequestableLayers;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::optional<WebMapLayer> WebMapService::getLayer(std::string const& name) const {
+  std::vector<WebMapLayer> layers = getLayers();
+  auto                     layer  = std::find_if(
+      layers.begin(), layers.end(), [name](WebMapLayer l) { return l.getName() == name; });
```
JonasGilg, line 71:
```suggestion
      layers.begin(), layers.end(), [name](WebMapLayer const& l) { return l.getName() == name; });
```

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+  if (layer == layers.end()) {
+    return {};
+  }
+
+  return *layer;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+bool WebMapService::isFormatSupported(std::string format) const {
```
JonasGilg, line 81:
```suggestion
bool WebMapService::isFormatSupported(std::string const& format) const {
```

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+        return {};
+      }
+
+      try {
+        WebMapExceptionReport e(resDoc);
+        if (e.getExceptions().size() == 1 &&
+            e.getExceptions()[0].getCode() == WebMapException::Code::eCurrentUpdateSequence) {
+          // Cache is up to date
+          return std::make_tuple(capabilitiesString, cacheDoc, false);
+        } else {
```
JonasGilg, line 245: You can remove the else after the return and save one indentation.

plugins/csp-wms-overlays/src/WebMapService.cpp
```diff
@@ -0,0 +1,354 @@
...
+          logger().warn("WMS Exception occurred while checking cache validity for '{}': '{}'!",
+              mUrl, e.what());
+          // Cache is not up to date, and an exception occured
```
JonasGilg, line 248:
```suggestion
          // Cache is not up to date, and an exception occurred
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+#define STB_IMAGE_WRITE_IMPLEMENTATION
+#include <stb_image.h>
+#include <stb_image_write.h>
+
+namespace csp::wmsoverlays {
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+WebMapTextureLoader::WebMapTextureLoader()
+    : mThreadPool(32) {
```
JonasGilg, line 33: Why use a fixed number of threads? Would a number relative to the number of cores be better?

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+WebMapTextureLoader::~WebMapTextureLoader() {
```
JonasGilg, line 38: You can remove empty destructors and just define them in the header with `= default`.

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  std::optional<WebMapTexture> texture = loadTextureFromStream(textureStream.value());
+  return texture;
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::optional<std::stringstream> WebMapTextureLoader::requestTexture(WebMapService const& wms,
+    WebMapLayer const& layer, Request const& wmsrequest, std::string const& mapCache) {
```
JonasGilg, line 81: The `mapCache` parameter is never used in this function.

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+    try {
+      request.perform();
+    } catch (std::exception& e) {
+      logger().warn("Failed to perform WMS request '{}': '{}'!", url, e.what());
+      continue;
+    }
+
+    std::string contentType = curlpp::Info<CURLINFO_CONTENT_TYPE, std::string>::get(request);
+    // Remove suffix and parameter from content type
+    size_t suffixPos    = contentType.find("+");
```
JonasGilg, line 110:
```suggestion
    size_t suffixPos    = contentType.find('+');
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+    size_t parameterPos = contentType.find(";");
```
JonasGilg, line 111:
```suggestion
    size_t parameterPos = contentType.find(';');
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::optional<WebMapTexture> WebMapTextureLoader::loadTextureFromFile(std::string const& fileName) {
+  int width, height, bpp;
+  int channels = 4;
+
+  unsigned char* pixels = stbi_load(fileName.c_str(), &width, &height, &bpp, channels);
+
+  if (!pixels) {
+    logger().warn("Failed to load '{}' with stbi!", fileName.c_str());
```
JonasGilg, line 199:
```suggestion
    logger().warn("Failed to load '{}' with stbi!", fileName);
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::optional<WebMapTexture> WebMapTextureLoader::loadTextureFromStream(
+    std::stringstream const& stream) {
+  int width, height, bpp;
+  int channels = 4;
+
+  unsigned char* pixels = stbi_load_from_memory((unsigned char*)stream.str().data(),
+      (int)stream.str().size(), &width, &height, &bpp, channels);
```
JonasGilg, line 215:
```suggestion
  unsigned char* pixels = stbi_load_from_memory(reinterpret_cast<unsigned char*>(stream.str().data()),
      static_cast<int>(stream.str().size()), &width, &height, &bpp, channels);
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+  if (request.mTime.has_value()) {
+    std::string       year;
+    std::stringstream time_stringstream(request.mTime.value());
+
+    // Create dir for year.
+    std::getline(time_stringstream, year, '-');
+    cacheDir << year << "/";
+  }
+
+  if (request.mStyle != "") {
```
JonasGilg, line 253:
```suggestion
  if (!request.mStyle.empty()) {
```

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
```diff
@@ -0,0 +1,358 @@
...
+      width = std::min(
+          request.mMaxSize, wms.getSettings().mMaxWidth.value_or(std::numeric_limits<int>::max()));
+    }
+  }
+
+  if (width.has_value() && !height.has_value()) {
+    height = (int)((double)width.value() / aspect);
+  } else if (height.has_value() && !width.has_value()) {
+    width = (int)((double)height.value() * aspect);
+  }
```
JonasGilg, line 331:
```suggestion
  if (width.has_value() && !height.has_value()) {
    height = static_cast<int>(static_cast<double>(width.value()) / aspect);
  } else if (height.has_value() && !width.has_value()) {
    width = static_cast<int>(static_cast<double>(height.value()) * aspect);
  }
```

plugins/csp-wms-overlays/src/utils.cpp
```diff
@@ -0,0 +1,331 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+namespace utils {
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+std::string timeToString(std::string const& format, boost::posix_time::ptime time) {
+  std::stringstream sstr;
+  // Delete is called by the std::locale constructor.
+  auto facet = new boost::posix_time::time_facet();
```
JonasGilg, line 31:
```suggestion
  auto* facet = new boost::posix_time::time_facet();
```

plugins/csp-wms-overlays/src/utils.cpp
```diff
@@ -0,0 +1,331 @@
...
+    std::regex r("P([[:d:]]+Y)?([[:d:]]+M)?([[:d:]]+D)?T([[:d:]]+H)?([[:d:]]+M)?([[:d:]]+S|[[:d:]]+"
+                 "\\.[[:d:]]+S)?");
+    matchDuration(isoString, r, duration);
+  }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+void convertIsoDate(std::string& date, boost::posix_time::ptime& time) {
+  if (date.find("T") != std::string::npos) {
```
JonasGilg, line 93:
```suggestion
  if (date.find('T') != std::string::npos) {
```

plugins/csp-wms-overlays/src/utils.cpp
```diff
@@ -0,0 +1,331 @@
...
+    // Remove timezone from date string.
+    // For now UTC offsets are not supported and will be ignored.
+    date = std::regex_replace(date, std::regex("(Z$|[+\\-][0-9]{2}:?([0-9]{2})?$)"), "");
+  }
+
+  date.erase(
+      std::remove_if(date.begin(), date.end(), [](unsigned char x) { return std::ispunct(x); }),
+      date.end());
+
+  std::size_t pos        = date.find("T");
```
JonasGilg, line 108:
```suggestion
  std::size_t pos        = date.find('T');
```

plugins/csp-wms-overlays/src/utils.cpp
```diff
@@ -0,0 +1,331 @@
...
+        }
+        break;
+      }
+    }
+    if (cs::utils::contains(tmp.mFormat, "T")) {
+      // Time parts are present, append 'Z'
+      tmp.mFormat.append("Z");
+    }
+
+    if (endDate == "") {
```
JonasGilg, line 174:
```suggestion
    if (endDate.empty()) {
```

plugins/csp-wms-overlays/src/utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+    boost::posix_time::ptime time, Duration duration, int multiplier = 1);
+
+/// Tries to get the value contained in a XML element.
+/// Starts at baseElement and then descends into the children given as childPath.
+/// The return value is empty if the requested element is not present.
+template <typename T>
+std::optional<T> getElementValue(
+    VistaXML::TiXmlElement* baseElement, std::vector<std::string> const& childPath = {}) {
+  VistaXML::TiXmlHandle elementHandle(baseElement);
+  for (std::string child : childPath) {
```
JonasGilg, line 115:
```suggestion
  for (std::string const& child : childPath) {
```

JonasGilg (changes requested):
Very nice work! This is a great addition to CosmoScout and definitely a default plugin that has something to show for itself to the average user.

I had a short look at the code and it already looks very nice. The documentation is excellent and the overall code quality is nice too. 

There are some minor things that I caught. Maybe you can use clang-tidy to check the code also? In pull requests of this size many small things can hide.

Schneegans:
Currently https://gitlab.com/libtiff/libtiff is not available. See https://github.com/cosmoscout/cosmoscout-vr/pull/244.

plugins/csp-wms-overlays/src/utils.hpp
Schneegans: Looks great! Thank you so much for your work on this!

plugins/csp-wms-overlays/src/WebMapTextureLoader.cpp
TheFridge76: As the threads are mostly IO bound, I think the optimal amount of threads is not really dependent on the number of cores. When requesting new textures with curl I doubt the CPU will be a bottleneck, so it shouldn't be necessary to reduce the thread count on processors with fewer cores.
I agree though, that the magic number is not really an elegant solution. However, even when creating a number of threads relative to the core count I would propose running a fixed amount >1 of threads on each core, which again would introduce another magic number. What do you think about this?

Schneegans:
I played around a bit and it seems pretty stable now. I applied the automatic fixes from `clang-format` and `clang-tidy`. The only thing I would suggest is to remove `"https://svs.gsfc.nasa.gov/cgi-bin/wms"` from the default config as it causes several warnings during startup which is not-so-good for the user experience.

PR #238, opened 2021-02-10
Safer Signals and Slots

JonasGilg:
Added safeguards, that prevent a crash, when a slot is disconnected while the signal is emitting all its slots.

The slots that are disconnected while emitting will be stored and disconnected after all other slots have been called.

### DIFF
--- a/src/cs-utils/Signal.hpp
+++ b/src/cs-utils/Signal.hpp
@@ -51,28 +51,46 @@ class Signal {
 
   /// Disconnects a previously connected function.
   void disconnect(int id) const {
-    mSlots.erase(id);
+    if (mIsIterating) {
+      mSlotsToDisconnect.push_back(id);
+    } else {
+      mSlots.erase(id);
+    }
   }
 
   /// Disconnects all previously connected functions.
   void disconnectAll() const {
-    mSlots.clear();
+    if (mIsIterating) {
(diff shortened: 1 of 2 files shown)


src/cs-utils/Signal.hpp
```diff
@@ -92,8 +110,23 @@ class Signal {
...
 
  private:
+  void postEmitCleanUp() {
+    if (mDisconnectAllRequested) {
+      mSlots.clear();
+      mDisconnectAllRequested = false;
+    } else {
+      for (int id : mSlotsToDisconnect) {
+        mSlots.erase(id);
+      }
```
Schneegans, line 120: Shouldn't this be cleared? E.g. like this:

```suggestion
      }
      mSlotsToDisconnect.clear();
```

Schneegans (changes requested):
Looks good - except for the one thing I mentioned below.

If you checked whether `mIsIterating` is already `true` at the beginning of the `emit*` methods, you could catch recursive emits which are also sources of crashs, I think.

JonasGilg:
Recursive calls to emit now get ignored and a warning is logged. Connecting new slots while emitting is not a problem. The standard assures the iterator is not invalidated while inserting into a map.

Schneegans (approved):

PR #239, opened 2021-02-17
Advanced Glare

Schneegans:
This PR changes the following:
* Add a new perspective correct glare model. The existing glare is computed in screen space which makes the glaring appear "to be painted on the screen" rather than being circular around the light source. This is because circular glares on peripheral areas of the screen look like ellipses from the perspective of the user. To counter-balance this, the glaring must be computed asymmetrically.
* Tweak the existing glare to be more physically based. The glare point-spread function now closely resembles what has been proposed in the literature (e.g. https://pubmed.ncbi.nlm.nih.gov/14632612/)
* Remove the auto-glare option since the new models work regardless of the current exposure.
* Add options to control the quality of the glare.
* Limit the default maximum auto-exposure a bit so that HDR-stars look more similar to non-HDR-stars.
* Rename any occurrence of "Glow" to "Glare".

Note that the new glare is much more expensive to compute. To look good, the glare quality slider must be moved some steps to the right. For this reason I decided not to use the perspective correct glare per default. To see the differences, the field of view must be quite large. Here are some examples:

| Symmetric (fast) | Asymmetric (correct) |
| -------------- | ---------------  |
Screenshot from 2021-02-17 06-53-37 | Screenshot from 2021-02-17 06-53-21
Screenshot from 2021-02-17 07-50-38 | Screenshot from 2021-02-17 07-50-31


JonasGilg:
With the bi-cubic filter disabled the glare looks very blocky, even on the highest quality.
glare

I managed to get a really flickery glare. The auto adjustments seem to overcompensate.
glare

src/cs-graphics/GlareMipMap.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+      glMemoryBarrier(GL_SHADER_IMAGE_ACCESS_BARRIER_BIT);
+      glDispatchCompute(static_cast<uint32_t>(std::ceil(1.0 * width / 16)),
+          static_cast<uint32_t>(std::ceil(1.0 * height / 16)), 1);
+    }
+  }
+
+  glBindImageTexture(0, 0, 0, GL_FALSE, 0, GL_WRITE_ONLY, GL_RGBA32F);
+  glBindImageTexture(1, 0, 0, GL_FALSE, 0, GL_READ_ONLY, GL_RGBA32F);
+}
```
JonasGilg, line 444: Do we also need to unbind the texture ate location 2 or is it used immediately after?

JonasGilg (changes requested):

src/cs-graphics/HDRBuffer.cpp
```diff
@@ -105,10 +105,10 @@ void HDRBuffer::bind() {
     // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
     hdrBuffer.mLuminanceMipMap = new LuminanceMipMap(mMultiSamples, size[0], size[1]);
 
-    // Create glow mipmaps.
-    delete hdrBuffer.mGlowMipMap; // NOLINT(cppcoreguidelines-owning-memory)
+    // Create glare mipmaps.
+    delete hdrBuffer.mGlareMipMap; // NOLINT(cppcoreguidelines-owning-memory)
     // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
-    hdrBuffer.mGlowMipMap = new GlowMipMap(mMultiSamples, size[0], size[1]);
+    hdrBuffer.mGlareMipMap = new GlareMipMap(mMultiSamples, size[0], size[1]);
```
JonasGilg, line 111: This manual memory management seems dangerous. Not just here, but throughout the whole file.

I also can't find a **final** `delete` for this and some other manually managed pointers. Do they leak memory?

Schneegans:
> With the bi-cubic filter disabled the glare looks very blocky, even on the highest quality.

That's the reason for the existence of the bi-cubic filter. As the resolution of the glare mip-map is quite low, bi-cubic filtering is required to achieve a smooth glow. The quality-slider adjusts the amount of samples taken to generate the mip-map, the bi-cubic filtering is then later used for loop-ups in the mip-map. So you cannot compensate one with the other. The purpose of the checkbox is to allow for evaluation of the quality and performance impact of the filtering.

BTW, the performance impact is quite huge. If you have an idea for more performant look-ups, I would be very grateful!
 
> I managed to get a really flickery glare. The auto adjustments seem to overcompensate.

I guess that this has nothing to do with this PR. It's been the same before - can you try pausing the playback time? Flickering should be gone. The reason is that the light source (Earth in your example) only covers a handful of pixels and as it moves slowly, the light energy contributed to the scene changes significantly due to aliasing. I still have no good idea how to fix this - I think the only way would be to fade out the actual planet if it's smaller than a few pixels and replace it by a procedural blob like the stars. This would then have a well-defined size in screen space. But computing the brightness of the blob is not easy as it depends on the atmosphere, the current phase and so on.

src/cosmoscout/Application.cpp
```diff
@@ -1119,24 +1119,48 @@ void Application::registerGuiCallbacks() {
...
-  // Toggles auto-glow.
-  mGuiManager->getGui()->registerCallback("graphics.setEnableAutoGlow",
-      "If enabled, the glow amount is chosen based on the current exposure.",
-      std::function([this](bool enable) { mSettings->mGraphics.pEnableAutoGlow = enable; }));
-  mSettings->mGraphics.pEnableAutoGlow.connectAndTouch([this](bool enable) {
-    // We fire callbacks (last param) to make sure that the glow-slider is properly activated /
-    // deactivated.
-    mGuiManager->setCheckboxValue("graphics.setEnableAutoGlow", enable, true);
+  // Adjusts the amount of artificial glare in HDR mode. If auto-glare is enabled, we update the
+  // slider in the user interface to show the current value.
```
Schneegans, line 1123:
```suggestion
  // Adjusts the amount of artificial glare in HDR mode.
```

src/cs-graphics/GlareMipMap.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+      // Rotate the view vector to the current pixel several times around the rotation axis
+      // in order to sample the vicinity.
+      for (float i = 0; i < samples; ++i) {
+        float angle  = totalAngle * i / (samples-1) - totalAngle * 0.5;
+        float sigma  = totalAngle / MAX_LEVELS;
+        float weight = getGauss(sigma, angle);
+
+        // Compute the rotated sample position in screen space.
+        vec4 pos = uMatP * vec4(rotate(posViewSpace.xyz, rotAxis, angle*PI/180.0), 1.0);
```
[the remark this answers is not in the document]
Schneegans: Because the literature on glare perception always uses angles. I could change it - it may increase peformance ever-so-slightly.

src/cs-graphics/GlareMipMap.cpp
```diff
@@ -0,0 +1,448 @@
...
+    
+    return mat3(i*axis.x*axis.x + c,        i*axis.x*axis.y - axis.z*s, i*axis.z*axis.x + axis.y*s,
+                i*axis.x*axis.y + axis.z*s, i*axis.y*axis.y + c,        i*axis.y*axis.z - axis.x*s,
+                i*axis.z*axis.x - axis.y*s, i*axis.y*axis.z + axis.x*s, i*axis.z*axis.z + c);
+  }
+
+  // Rotates the given vector around a given axis.
+  vec3 rotate(vec3 v, vec3 axis, float angle) {
+    return rotationMatrix(axis, angle) * v;
+  }
```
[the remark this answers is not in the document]
Schneegans: I based the code on https://gist.github.com/yiwenl/3f804e80d0930e34a0b33359259b556c... now we know where this guy had it's code from :smile:. I just tested the one-liner and it makes no visual difference.

src/cs-graphics/GlareMipMap.cpp
```diff
@@ -0,0 +1,448 @@
...
+  // Create glare mipmap storage. The texture has half the size of the HDR buffer (rounded down) in
+  // both directions.
+  int iWidth  = mHDRBufferWidth / 2;
+  int iHeight = mHDRBufferHeight / 2;
+
+  // Compute the number of available mipmap levels.
+  mMaxLevels =
+      static_cast<int>(std::max(1.0, std::floor(std::log2(std::max(iWidth, iHeight))) + 1));
+
+  Bind();
```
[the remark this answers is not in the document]
Schneegans: The binding is overidden by the Bind() further below.

src/cs-graphics/GlareMipMap.cpp
```diff
@@ -0,0 +1,448 @@
...
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_LINEAR);
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+
+  glTexStorage2D(GL_TEXTURE_2D, mMaxLevels, GL_RGBA32F, iWidth, iHeight);
+
+  // Create storage for temporary glare target (this is used for the vertical blurring passes).
+  mTemporaryTarget->Bind();
```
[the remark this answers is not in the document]
Schneegans: I can do it. But it really does not matter much because anyone doing something on GL_TEXTURE_2D will re-bind it anyways.

src/cs-graphics/GlareMipMap.cpp
Schneegans: Same as above :smile:

src/cs-graphics/HDRBuffer.cpp
Schneegans: You're right. This is an relic and should be fixed.

JonasGilg (approved):
Looks good 😄

PR #240, opened 2021-02-25
Improved timings

Schneegans:
This PR significantly improves the frame timing capabilities of CosmoScout VR. Here are the changes:
* The `FrameTimings` now keep track of the number of currently active ranges. This allows us to determine the current nesting level of a timing range.
* The statistics UI element now shows the ranges as they were recorded for the last-but-one frame with the appropriate nesting level. There are tooltips for detailed information.
* A triple buffer is used for the timer queries, which improves performance.
* The recording-to-file has been adapted accordingly. It now creates a set of files, one for each nesting level per CPU or GPU.
* The `FrameTimings` class is now a real singleton, before it used some hidden static variables.

Screenshot from 2021-02-24 13-42-39

### DIFF
--- a/plugins/csp-timings/gui/js/csp-timings.js
+++ b/plugins/csp-timings/gui/js/csp-timings.js
@@ -60,7 +60,7 @@ class TimingsApi extends IApi {
 
     // Create the color hash object for coloring the timing ranges.
     if (typeof ColorHash !== 'undefined') {
-      this._colorHash = new ColorHash({lightness: 0.4, saturation: 0.3});
+      this._colorHash = new ColorHash({lightness: 0.3, saturation: 0.6});
     } else {
       console.error('Class \'ColorHash\' not defined.');
     }
--- a/plugins/csp-timings/gui/js/csp-timings.js
+++ b/plugins/csp-timings/gui/js/csp-timings.js
@@ -130,7 +130,7 @@ class TimingsApi extends IApi {
       // 1 2 5 10 20 50 100 200 500 1000 ....
       let int  = ~~(i / 3);
       let mod  = i % 3;
-      interval = (mod == 1 ? i * 2 : i * 2 + 1) * (int + 1);
+      interval = (mod === 1 ? i * 2 : i * 2 + 1) * (int + 1);
     }
 
     // Compute the number of required grid lines.
@@ -142,14 +142,14 @@ class TimingsApi extends IApi {
       // The last tick line will require a little bit of margin on the right hand side as the
       // maxTime will not be a full multiple of interval.
       let margin = "";
-      if (i == ticks) {
+      if (i === ticks) {
         let fullTickWidth = 100.0 / (ticks + 1);
         let remainder     = maxTime / interval - ticks;
         margin            = `style="margin-right: ${fullTickWidth * remainder}%"`;
       }
 
       // Every 5 ticks we add a major grid line.
-      if (i % 5 == 0) {
+      if (i % 5 === 0) {
         gridLines.innerHTML +=
             `<div class="tick major" ${margin}><span>${i * interval} ms</span></div>`
       } else {
--- a/plugins/csp-timings/gui/js/csp-timings.js
+++ b/plugins/csp-timings/gui/js/csp-timings.js
@@ -1,6 +1,11 @@
 /* global IApi, CosmoScout, ColorHash */
 
 /**
+ * This amount of frame data will be stored to slide through.
+ */
+const maxStoredFrames = 100;
+
+/**
  * Timings Api
  */
 class TimingsApi extends IApi {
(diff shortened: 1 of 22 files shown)


JonasGilg:
The UI for the Frame Timings now only displays the time for a single frame? I actually would also like a running average as we had before, since it gives a good overview of the performance, especially with more complex plugins, that might be dropping frames irregularly, like the trajectories plugin.

The current implementation is also great, since it gives really deep information about the rendered frame.

Maybe make it configurable in the UI. Have a slider for calculating the average for the last `n` frames, with `n = 1` for the current frame only. 

Or a slider to modify the running average ratio. I think previously we had a 5% to 95% ratio. Maybe you can set that with a slider and with a 100% to 0% ratio you get only the current frame.

plugins/csp-timings/gui/js/csp-timings.js
```diff
@@ -33,143 +17,146 @@ class TimingsApi extends IApi {
...
-    for (let i = 0; i < maxEntries; ++i) {
-      /* eslint-disable no-mixed-operators */
-      const widthGPU = maxWidth * this._values[i].avgTimeGPU / this._maxValue;
-      const widthCPU = maxWidth * this._values[i].avgTimeCPU / this._maxValue;
-      /* eslint-enable no-mixed-operators */
+      // For linearly increasing i, this results in these intervals:
+      // 1 2 5 10 20 50 100 200 500 1000 ....
+      let int  = ~~(i / 3);
+      let mod  = i % 3;
+      interval = (mod == 1 ? i * 2 : i * 2 + 1) * (int + 1);
```
JonasGilg, line 133:
```suggestion
      interval = (mod === 1 ? i * 2 : i * 2 + 1) * (int + 1);
```

plugins/csp-timings/gui/js/csp-timings.js
```diff
@@ -33,143 +17,146 @@ class TimingsApi extends IApi {
...
+    let ticks       = Math.floor(maxTime / interval);
+    const gridLines = document.createElement('template');
 
-      container.appendChild(item.content);
+    for (let i = 0; i <= ticks; i++) {
+
+      // The last tick line will require a little bit of margin on the right hand side as the
+      // maxTime will not be a full multiple of interval.
+      let margin = "";
+      if (i == ticks) {
```
JonasGilg, line 145:
```suggestion
      if (i === ticks) {
```

plugins/csp-timings/gui/js/csp-timings.js
```diff
@@ -33,143 +17,146 @@ class TimingsApi extends IApi {
...
+        let fullTickWidth = 100.0 / (ticks + 1);
+        let remainder     = maxTime / interval - ticks;
+        margin            = `style="margin-right: ${fullTickWidth * remainder}%"`;
+      }
+
+      // Every 5 ticks we add a major grid line.
+      if (i % 5 == 0) {
```
JonasGilg, line 152:
```suggestion
      if (i % 5 === 0) {
```

plugins/csp-timings/src/Plugin.cpp
```diff
@@ -117,96 +117,158 @@ void Plugin::init() {
...
-        uint64_t timeCPU(timing.second.mCPUTime);
-
-        uint64_t const thresholdNanos = 100000;
-        if (timeGPU > thresholdNanos || timeCPU > thresholdNanos) {
-          json += "\"" + timing.first + "\":[" + std::to_string(timeGPU) + "," +
-                  std::to_string(timeCPU) + "],";
+      auto toJSON = [](std::vector<std::vector<Range>> const& ranges) {
+        nlohmann::json json;
+
+        for (uint32_t i = 0; i < ranges.size(); ++i) {
```
JonasGilg, line 166: This can be a range based loop.

plugins/csp-timings/src/Plugin.cpp
```diff
@@ -117,96 +117,158 @@ void Plugin::init() {
...
-    // Get the frame timing information if either recording or the on-screen statistics are enabled.
-    std::unordered_map<std::string, cs::utils::FrameTimings::QueryResult> timings;
-    if (mEnableStatistics || mEnableRecording) {
-      timings = mFrameTimings->getCalculatedQueryResults();
+    // Only record ranges longer than 10 µs.
+    const uint32_t minTimeNanos    = 10000;
+    auto const&    ranges          = cs::utils::FrameTimings::get().getRanges();
+    uint32_t       maxNestingLevel = 0;
+
+    // Compute the maximum nesting level amongst all recoreded ranges.
```
JonasGilg, line 132:
```suggestion
    // Compute the maximum nesting level amongst all recorded ranges.
```

plugins/csp-timings/src/Plugin.cpp
```diff
@@ -117,96 +117,158 @@ void Plugin::init() {
...
-    for (auto const& frameTiming : mRecordedTimings) {
-      for (auto const& timing : frameTiming) {
-        timerNames.insert(timing.first);
+    std::string directory = "csp-timings/" + timeString;
+    cs::utils::filesystem::createDirectoryRecursively(
+        boost::filesystem::system_complete(directory));
+
+    // This stores a CSV file for each nesting level in the directory created above. The prefix will
+    // be prepended to the CSV file name.
+    auto saveRecording = [directory](std::string const&                          prefix,
```
JonasGilg, line 208:
```suggestion
    auto saveRecording = [&directory](std::string const&                          prefix,
```

JonasGilg (changes requested):

plugins/csp-timings/src/Plugin.cpp
Schneegans: Done.

Schneegans:
Implementing an average on this kind of data is close-to-impossible. The reason is that the number of ranges, their nesting level, the ordering - basically anything which is shown in the graph may completely change from one frame to another. Of course it looks oftentimes similar from one frame to the next but there is no guarantee... Implementing something which tries to match one of these datasets to the next will be really hard. 

That being said, I implemented another variant which is even better (at least in my opinion): We now have a slider in the statistics UI which lets you scroll through the last couple of frames! This way we can easily explore what took longer in the stuttering frames. What do you think?

JonasGilg:
> Implementing an average on this kind of data is close-to-impossible. The reason is that the number of ranges, their nesting level, the ordering - basically anything which is shown in the graph may completely change from one frame to another. Of course it looks oftentimes similar from one frame to the next but there is no guarantee... Implementing something which tries to match one of these datasets to the next will be really hard.
> 
> That being said, I implemented another variant which is even better (at least in my opinion): We now have a slider in the statistics UI which lets you scroll through the last couple of frames! This way we can easily explore what took longer in the stuttering frames. What do you think?

This looks great. I found a small bug though. When you scroll through the frames, before there is any data available I get the following error:
```
[E] cs-gui ............ Uncaught TypeError: Cannot read property '0' of undefined
```

plugins/csp-timings/gui/js/csp-timings.js
```diff
@@ -130,7 +188,7 @@ class TimingsApi extends IApi {
       // 1 2 5 10 20 50 100 200 500 1000 ....
       let int  = ~~(i / 3);
```
JonasGilg, line 189:
```suggestion
      let int  = Math.floor(i / 3);
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #241, opened 2021-03-01
Bookmark fixes

Schneegans:
This fixes the following:
* Bookmarks with orientation did not set the correct orientation.
* It wasn't possible to set bookmarks relative to the Solar System's Barycenter because Jupiter kept being the active body even if zoomed out very far.

### DIFF
--- a/src/cs-core/SolarSystem.cpp
+++ b/src/cs-core/SolarSystem.cpp
@@ -395,14 +395,14 @@ void SolarSystem::updateObserverFrame() {
   }
 
   // If currently no observer animation is in progress, we change the pActiveBody accordingly. This
-  // may be null if we are very far awy from any object.
+  // may be null if we are very far away from any object.
   if (!mObserver.isAnimationInProgress()) {
     pActiveBody = activeBody;
 
     std::string sCenter = "Solar System Barycenter";
     std::string sFrame  = "J2000";
 
-    // We change frame and center if there is a object with weight larger than mLockWeight
+    // We change frame and center if there is an object with weight larger than mLockWeight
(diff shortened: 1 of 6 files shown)


src/cs-core/SolarSystem.cpp
```diff
@@ -393,23 +394,27 @@ void SolarSystem::updateObserverFrame() {
...
   }
 
-  // We change frame and center if there is a object with weight larger than mLockWeight
-  // and mTrackWeight.
-  if (activeBody) {
-    if (!mObserver.isAnimationInProgress()) {
-      std::string sCenter = "Solar System Barycenter";
-      std::string sFrame  = "J2000";
+  // If currently no observer animation is in progress, we change the pActiveBody accordingly. This
+  // may be null if we are very far awy from any object.
```
JonasGilg, line 398:
```suggestion
  // may be null if we are very far away from any object.
```

src/cs-core/SolarSystem.cpp
```diff
@@ -393,23 +394,27 @@ void SolarSystem::updateObserverFrame() {
...
+  if (!mObserver.isAnimationInProgress()) {
+    pActiveBody = activeBody;
 
+    std::string sCenter = "Solar System Barycenter";
+    std::string sFrame  = "J2000";
+
+    // We change frame and center if there is a object with weight larger than mLockWeight
```
JonasGilg, line 405:
```suggestion
    // We change frame and center if there is an object with weight larger than mLockWeight
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #249, opened 2021-03-26
Add Changelog

Schneegans:
Did I forget something?


JonasGilg (approved):
What do you think about a rolling changelog, where each merge request adds a line to the changelog before merging.

Schneegans:
Good idea! So we should add the 1.5.0 directly to the changelog once 1.4.0 is released.

PR #251, opened 2021-03-30
Fix CI

Schneegans:
With the boost / cmake versions installed on the GitHub actions runners, `Boost_LIBRARIES` does not contain anything. Therefore we have to use the individual boost library variables to ensure everything gets installed properly.


JonasGilg (approved):
As a side note do we actually make use of boost::chrono?

CMakeLists.txt
```diff
@@ -173,7 +176,16 @@ install(
...
-foreach(BOOST_LIB ${Boost_LIBRARIES})
+foreach(BOOST_LIB 
+  ${Boost_CHRONO_LIBRARY_DEBUG}
+  ${Boost_CHRONO_LIBRARY_RELEASE}
+  ${Boost_DATE_TIME_LIBRARY_DEBUG}
+  ${Boost_DATE_TIME_LIBRARY_RELEASE}
+  ${Boost_FILESYSTEM_LIBRARY_RELEASE}
+  ${Boost_FILESYSTEM_LIBRARY_DEBUG}
+  ${Boost_SYSTEM_LIBRARY_RELEASE}
+  ${Boost_SYSTEM_LIBRARY_RELEASE}
```
JonasGilg, line 187:
```suggestion
  ${Boost_SYSTEM_LIBRARY_DEBUG}
```

JonasGilg (changes requested):

Schneegans:
Good catch about the typo. I think we do not use it directly but it's a dependency of boost::date_time, I think.

JonasGilg (approved):

PR #252, opened 2021-03-31
Virtual Guide Plugin

JonasGilg:
This awesome plugin adds a super helpful guide to CosmoScout. He will be subtle and hints the user in the right direction, when he needs it.

Try it out!

Guide

Currently the guide can greet you, show you the orbit controls when rotating the planet and show you the bookmarks tab when flying away from Earth.


Schneegans:
This looks like a very important contribution! CosmoScout VR has definitely some deficiencies when it comes to user experience. However, I cannot fully appreciate it as I get some errors when loading the plugin:

```
[E] cs-gui ............ Failed to open gui resource: Cannot open file '../share/resources/gui/third-party/agents/Clippy/agent.js'!
[E] cs-gui ............ Failed to open gui resource: Cannot open file '../share/resources/gui/third-party/agents/Clippy/sounds-mp3.js'!
[E] cs-gui ............ Failed to open gui resource: Cannot open file '../share/resources/gui/third-party/agents/Clippy/map.png'!
```

Can you help me to resolve these?

Schneegans:
This is really awesome! This must have been some work on your side :). I think we should merge this - there's only one tiny thing I would ask you before: Could you ask Microsoft to hand over the copyright for those characters? If they decline, we have to render our own character. Looking at the source of Clippy.js, it shouldn't be too difficult to add a custom character, let's say a small curiosity rover. What do you think?

JonasGilg:
> This is really awesome! This must have been some work on your side :). I think we should merge this - there's only one tiny thing I would ask you before: Could you ask Microsoft to hand over the copyright for those characters? If they decline, we have to render our own character. Looking at the source of Clippy.js, it shouldn't be too difficult to add a custom character, let's say a small curiosity rover. What do you think?

This is a good idea. I'll ask them. I think in the long run we should create our own character anyways. Sadly I am very bad at graphic design and have no experience with animation.

PR #254, opened 2021-04-02
PowerShell script to generate plugins

JonasGilg:
This was a fun project by me to learn more about PowerShell. It creates a plugin in the plugins folder with the given name.

## Description

This function creates a new CosmoScout VR plugin with the given name. You can also choose a
Material Icon that will be set for your plugin in the UI. It will generate all relevant files for
your Plugin:
- README.md
- CMakeLists.txt
- src/Plugin.hpp
- src/Plugin.cpp
- src/logger.hpp
- src/logger.cpp
  
It will also generate the following files by default, but you can disable them, if you don't
want them:
- gui/js/csp-<plugin-name>.js
- gui/plugin-name-tab.html
- gui/plugin-name-settings.html
  
#### Parameter: `-Name` 
The plugin name. Just write it out in natural language. Don't use CamelCase, KebabCase, 
PascalCase or SnakeCase. The script will convert it to the correct case for each occurance.
  
#### Parameter: `-Icon`
The name of the Material Icon that will be shown in the GUI for this plugin. You can find a
list of icons to choose from here: https://fonts.google.com/icons?selected=Material+Icons

#### Parameter: `-WithJSApi`
This will create a JavaScript Api class in the plugins gui folder and create the appropriate
code in the Plugin.cpp file to load and register the class with the CosmoScout Api.

#### Parameter: `-WithSidebarTab`
This will generate an HTML file, where you can write your sidebar tab content. The script will
also create the appropriate code in the Plugin.cpp file to load and unload the sidebar tab.

#### Parameter: `-WithSettingsTab`
This will generate an HTML file, where you can write your settings tab content. The script will
also create the appropriate code in the Plugin.cpp file to load and unload the settings tab.

## Examples:
```
PS> .\New-Plugin -Name "Hello World" -Icon "hail"

PS> .\New-Plugin "Hello World" "hail"

PS> .\New-Plugin "Hello World" -WithJSApi $false -WithSidebarTab $false -WithSettingsTab $false
```


Schneegans:
This is great! There are only a few things I could ask for:
* A Linux version of this (but this is maybe out-of-scope and could be handled in a separate PR)
* A hint to this script in https://github.com/cosmoscout/cosmoscout-vr/blob/develop/docs/plugins.md
* Is the file name of this script somehow fixed? Else I would suggest to name it more similar to our other scripts. Like `create_new_plugin.ps1`.

But anyways, that's a cool contribution.

Schneegans:
Btw, I am trying to fix the CI in #255.

JonasGilg:
> * A Linux version of this (but this is maybe out-of-scope and could be handled in a separate PR)

PowerShell also [works under Linux](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-7.1), although the user has to install it.

> * A hint to this script in https://github.com/cosmoscout/cosmoscout-vr/blob/develop/docs/plugins.md

I will do that.

> * Is the file name of this script somehow fixed? Else I would suggest to name it more similar to our other scripts. Like `create_new_plugin.ps1`.

I oriented myself on the official [PowerShell naming conventions](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.1), although we could deviate from that.

Maybe we should put utilities like this in a folder like `tools`? They don't get used regularly and clutter the root folder.

Schneegans (approved):

PR #256, opened 2021-04-09
Feature/windows scripts

JonasGilg:
This creates clang-format and clang-tidy PowerShell scripts.


Schneegans:
I'll test those as soon as possible. In the meantime - what do you think about moving all those formatting scripts to the tools directory?

JonasGilg:
> [...] - what do you think about moving all those formatting scripts to the tools directory?

That's a good idea.

Schneegans:
I just tested this and it seems to work properly. Just two observations:
* The `CMAKE_EXPORT_COMPILE_COMMANDS` only works with ninja. Therefore the clang-tidy script only works with this generator as well. Should we point this out somewhere? At least in the comments in the file?
* Not really related to this PR, but I just noticed that the new-plugin-script behaves kind of weird when no parameter is passed. Maybe it could either ask for the plugin name or print a warning / help message.

What do you think?

Schneegans:
Btw, the comment percentage check must fail for this PR as the script got moved. So it cannot find it in the base branch.

JonasGilg:
> * The `CMAKE_EXPORT_COMPILE_COMMANDS` only works with ninja. Therefore the clang-tidy script only works with this generator as well. Should we point this out somewhere? At least in the comments in the file?

Yeah I'll add that to the description.

> * Not really related to this PR, but I just noticed that the new-plugin-script behaves kind of weird when no parameter is passed. Maybe it could either ask for the plugin name or print a warning / help message.

When I call the New-Plugin.ps1 without parameters I get a prompt asking for the name by default:
```powershell
PS C:\C++\cs-debug> .\tools\New-Plugin.ps1

cmdlet New-Plugin.ps1 at command pipeline position 1
Supply values for the following parameters:
Name[0]:
```

Schneegans:
> When I call the New-Plugin.ps1 without parameters I get a prompt asking for the name by default:
> 
> ```powershell
> PS C:\C++\cs-debug> .\tools\New-Plugin.ps1
> 
> cmdlet New-Plugin.ps1 at command pipeline position 1
> Supply values for the following parameters:
> Name[0]:
> ```

Yes, it does, but if I enter a name, it keeps asking for other paramters called `Name`...?

JonasGilg:
> > When I call the New-Plugin.ps1 without parameters I get a prompt asking for the name by default:
> > ```powershell
> > PS C:\C++\cs-debug> .\tools\New-Plugin.ps1
> > 
> > cmdlet New-Plugin.ps1 at command pipeline position 1
> > Supply values for the following parameters:
> > Name[0]:
> > ```
> 
> Yes, it does, but if I enter a name, it keeps asking for other paramters called `Name`...?

Ah, I see. I think I fixed that with the last commit. I accidentally specified the `Name` parameter as an array of strings.

Schneegans (approved):

Schneegans:
It works now!

PR #264, opened 2021-09-07
VR Accessibility Improvements

Schneegans:
This provides various fixes and improvements for the `csp-vr-accessibilty` plugin. The most notable features are:

* The color and depth buffer not copied twice a frame anymore.
* Plenty of uniforms where not need actually.
* The grid now uses additive blending which fades much nicer in the distance.
* The grid is now slightly visible even if blocked by geometry. This makes it much more useful if one is standing on the ground.
* The grid scaling parameters are now independent from each other.
* The fade animation logic of the vignette has been simplified (no animation tracker anymore).
* SPICE is not used anymore to get the local time.
* The vignette settings use now two-handle sliders.
* The settings layout has been improved.
* The sidebar settings use a proper icon now.
* The dynamic radii are now faded out smoothly.
* The dynamic radii now use the fade duration property.


plugins/csp-vr-accessibility/README.md
```diff
@@ -41,7 +39,7 @@ This plugin can be enabled with the following configuration in your `settings.js
 }
 ```
 
-Most of the configuration (all options, except for Grid's `falloff` and `texture`) is also available at runtime in the "VR Accessibility" tab in the settings menu.
+Most of the configuration (all options, except for Grid's `offset` and `texture`) is also available at runtime in the "VR Accessibility" tab in the settings menu.
```
NXXR, line 42: shouldn't is be extent instead of offset?
the offset is the transform node, and should be modifiable at runtime.
only the Texture and size of the Quad is only modifiable within the settings.json and therefore need to reload the plugin to be modified.

NXXR (approved):

PR #285, opened 2022-06-03
Controller Support

NXXR:
## Changelog:
- Adds VRPN to externals
- Builds VRPN Server and config (for Original Xbox360 Controllers) into the release/bin folder
- Starts VRPN Server instance when starting cosmoscout 
  - controller has to be plugged in before starting cosmoscout
  - hot-switching does not work (if the controller looses connection/is unplugged cosmoscout needs to be restarted)
- Adds Vista DFN Control Scheme for controllers
- Adds Vista DFN to use mouse and controller, picking the latest position of either device
  - controller and mouse have their own, independent positions currently
- Adds Markdown [Mermaid diagrams](https://mermaid-js.github.io/mermaid/#/n00b-gettingStarted) as documentation for the controller DFNs in a doc folder at `config/base/vista/xml`

## Things to consider that are *not* in this PR:
- Add/Display a custom in-software cursor for the controller mouse to be visible
  - <details>
      <summary>Changes for Mouseray workaround:</summary>
      in `config/base/scene/simple_desktop.json` add:

      ```json
      "enableMouseRay": true,
      ```
      to top level settings (`startDate` etc.)
      
      ---
      in `src/cs-graphics/MouseRay.cpp` change/add:

      ```c++
      mRayTransform->SetScale(0.005F, 0.005F, 30.F);
      mRayTransform->SetTranslation(0.F, 0.F, -1.F);
      ```
      at ll. 86ff
    </details>
- Add more controller options to the config and DFNs
- Look into the Mouse-Controller-Negotioator to check if the controller can use the latest mouse position for the curor movement instead of storing its own position
- **VRPN Server is provided as an `.exe` and is only added to the batch scripts and is therefore currently only available in the windows builds.**


Schneegans:
With #360, is this branch now obsolete? What do you think, @JonasGilg?

NXXR:
As far as I know, Jonas got all he needed out of this.
And since the SDL Contorller Support seems to be working reasonably well (pending some more tests) this seems obsolete to me.

PR #306, opened 2022-10-06
Cleanup GUI manager interface (and fix "removing undefined child of node"-error)

Schneegans:
I think that this is going to be the last PR before we can publish 1.6.0!

With this PR, the interface of the `GuiManager` is streamlined. Before, HTML templates could be added via `addHtmlToGui()` but had to be removed with an explicit JavaScript call. The same was true for adding and removing stylesheets. Here are all the changes:
  * `addScriptToGui()` has been removed. It was never used and is actually identical to `getGui()->executeJavascript()`
  * `addScriptToGuiFromJS()` has been renamed to `executeJavascriptFile` as this describes more clearly what the method actually does.
  * `addHtmlToGui()` has been renamed to `addTemplate()`, as this is what the method actually does.
  * A corresponding `removeTemplate()` has been added.
  * `addCssToGui()` has been renamed to `addCSS()`.
  * A corresponding `removeCSS()` has been added.

All plugins now use this updated API. Furthermore, some plugins have been simplified by moving their HTML templates to the sidebar HTML files. This reduces the number of required files and a few lines of code for registering and unregistering the templates.

Edit: I also explicitely download `clang-format-11` for the formatting check. This way it's easier to reproduce the formatting locally.


src/cs-core/GuiManager.hpp
```diff
@@ -166,27 +167,24 @@ class CS_CORE_EXPORT GuiManager {
...
   /// This can be used to initialize the DOM elements added to the sidebar with the methods above.
   ///
   /// @param jsFile The javascript file that contains the source code.
-  void addScriptToGuiFromJS(std::string const& jsFile);
+  void executeJavascriptFile(std::string const& jsFile);
 
-  /// Append HTML to the body.
-  /// The src content will be wrapped in a template element.
+  /// Append an HTML template to the main GUI element.
+  /// The content of the given HTML file can then later be instantiated by the JavaScrip method
```
JonasGilg, line 176:
```suggestion
  /// The content of the given HTML file can then later be instantiated by the JavaScript method
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #312, opened 2022-10-28
Update all references to "master" and "develop" by "main"

Schneegans:
I hope that I got everything...


JonasGilg (approved):
I think you got everything. Once Vista and OpenSG are changed aswell this needs to be adressed here too. We could do it in alternate PR though,

PR #316, opened 2022-11-29
Capture in meters

Schneegans:
This fixes a regression which causes the depth images which can be captured with the  `csp-web-api` plugin not to be in meters anymore. This was caused by our changed projection matrix.

### DIFF
--- a/plugins/csp-web-api/src/Plugin.cpp
+++ b/plugins/csp-web-api/src/Plugin.cpp
@@ -421,7 +421,7 @@ void Plugin::update() {
         // Capture the depth component.
         std::vector<float> capture(mCaptureWidth * mCaptureHeight);
         glReadPixels(
-            0, 0, mCaptureWidth, mCaptureHeight, GL_DEPTH_COMPONENT, GL_FLOAT, &capture[0]);
+            0, 0, mCaptureWidth, mCaptureHeight, GL_DEPTH_COMPONENT, GL_FLOAT, capture.data());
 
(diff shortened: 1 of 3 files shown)


plugins/csp-web-api/src/Plugin.cpp
```diff
@@ -423,6 +424,21 @@ void Plugin::update() {
...
+          std::array<GLfloat, 16> glMatP{};
+          glGetFloatv(GL_PROJECTION_MATRIX, glMatP.data());
+          auto      matInvP = glm::inverse(glm::make_mat4x4(glMatP.data()));
+          glm::vec2 pixel(1.f / mCaptureWidth, 1.f / mCaptureHeight);
+
+          for (size_t i(0); i < capture.size(); ++i) {
+            auto coords = glm::vec2(i % mCaptureWidth, i / mCaptureWidth) * pixel + 0.5f * pixel;
+            auto pos    = matInvP * glm::vec4(2.f * coords - 1.f, 2.f * capture[i] - 1.f, 1.f);
+
+            float dist = glm::length(pos.xyz() / pos.w) * mSolarSystem->getObserver().getScale();
```
JonasGilg, line 438:
```suggestion
            float dist = static_cast<float>(
                (glm::length(pos.xyz() / pos.w) * mSolarSystem->getObserver().getScale()));
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #318, opened 2022-12-02
Use ccache for CI

### LINKED ISSUE #299: Use buildcache instead of clcache
## The Problem
`clcache` causes build errors on Windows and is [not maintained anymore](https://github.com/frerich/clcache).

## The Solution
Use [buildcache](https://github.com/mbitsnbites/buildcache) instead.
We could also consider to use this on Linux as well.

Schneegans:
This may resolve #299

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -198,6 +198,7 @@ jobs:
       contains(github.event.head_commit.message, '[run-ci]')
     env:
       BOOST_ROOT_1_72_0: C:\hostedtoolcache\windows\Boost\1.72.0\x86_64
+      BUILDCACHE_IMPERSONATE: cl.exe
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


Schneegans:
Some preliminary results. The numbers are extracted from various builds and should only be considered as rough estimates. The `clcache` numbers I cannot really find anymore as the logs for those runs do not exist anymore. But I found some ancient runs which took about this time.

build type | without caching | buildcache | ccache | clcache
-- | -- | -- | -- | --
gcc | 24 min | 8 min | 10 min | -
clang | 28 min | 8 min | 7 min | -
msvc | > 60 min | | - | (?)
msvc (pch + unity build)| 34 min |  | - | 28 min (?)

Schneegans:
A significant part of the time the windows build takes is for downloading boost. Maybe we should consider setting up a container for building which already includes boost. This could speed up the process by more than five minutes.

JonasGilg (approved):

PR #325, opened 2023-01-26
More queries

Schneegans:
This MR adds support for sample queries and primitive queries to the `FrameTimings` class. As a consequence, the class has been renamed to `FrameStats`.


resources/gui/js/apis/utils.js
```diff
@@ -85,21 +85,67 @@ class UtilsApi extends IApi {
...
+
+    const abs = Math.abs(value);
+    value     = Number(value);
+
+    if (abs < 1e-9) {
+      return this.formatNumber(value * 1e9) + "μ";
+    }
+    if (abs < 1e-6) {
+      return this.formatNumber(value * 1e6) + "n";
+    }
```
JonasGilg, line 125: If I am not mistaken nano and micro are switched here:

```suggestion
    if (abs < 1e-9) {
      return this.formatNumber(value * 1e9) + "n";
    }
    if (abs < 1e-6) {
      return this.formatNumber(value * 1e6) + "μ";
    }
```

src/cs-utils/FrameStats.cpp
```diff
@@ -0,0 +1,464 @@
...
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+void FrameStats::endPrimitivesQuery(int32_t id) {
+
+  // Only attempt to end the counting if pEnableMeasurements is set to true.
+  if (pEnableMeasurements.get()) {
+    mQueryPools.at(mCurrentQueryPool)->endPrimitivesQuery(id);
+  }
+}
```
JonasGilg, line 175: Does the case exist, where:
1. A query is started
2. The measurements are being disabled
3. The query isn't ended

or does the control flow not allow this scenario?

JonasGilg (changes requested):

JonasGilg (approved):

PR #327, opened 2023-01-26
Enable OpenGL logging. Prints errors, warnings, infos, etc.

Schneegans:


src/cs-graphics/SetupGLNode.cpp
```diff
@@ -27,6 +77,9 @@ bool SetupGLNode::Do() {
   glEnable(GL_CULL_FACE);
   glCullFace(GL_BACK);
 
+  // Attach the debug callback to print the messages
+  glDebugMessageCallback(MessageCallback, static_cast<void*>(mSettings.get()));
```
JonasGilg, line 81: Isn't the `Do` function part of the frame loop and called every frame? This function has to be called only once at startup.

src/cs-graphics/SetupGLNode.cpp
```diff
@@ -15,6 +16,55 @@ namespace cs::graphics {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+void GLAPIENTRY MessageCallback(GLenum source, GLenum type, GLuint id, GLenum severity,
```
JonasGilg, line 19: You can use `else if` or early `return` to avoid unneccesary boolean evaluations.

JonasGilg (changes requested):
There was already some kind of OpenGL logging here: https://github.com/cosmoscout/cosmoscout-vr/blob/main/src/cs-core/GraphicsEngine.cpp#L262
Although you had to enable it in code.

I also don't understand the functionality of the `SetupGLNode` class. This needs to be documented better. Isn't it better to do this in the GraphicsManager, since it will log all OpenGL messages?

Schneegans:
You are completely right! I actually wasn't aware of that method in the `GraphicsEngine`. :rofl: 

I'll fix this ASAP.

JonasGilg (approved):

Issue #331, opened 2023-02-21
"Tile loading failed" due to incomplete check


michaelschleiss:
The variable `fail` will evaulate to True in my case because contentType is "image/png; mode=8bit" instead of plain "image/png". 

https://github.com/cosmoscout/cosmoscout-vr/blob/ff5d1d3280c6c0a9e19c2fa466de2fe8aa702083/plugins/csp-lod-bodies/src/TileSourceWebMapService.cpp#L406-L407

I followed the Documentation [here](https://github.com/cosmoscout/cosmoscout-vr/tree/main/plugins/csp-lod-bodies) to setup the Map Server.

Schneegans:
Hey there, thanks for the report! I guess that the easiest fix would be to check whether the content type contains `image/png` or `image/tiff`. Like this:

```cpp
fail = !cs::utils::contains(contentType, "image/png") && !cs::utils::contains(contentType, "image/tiff");
```
I assume that you compiled CosmoScout VR from source, so you could directly try this.
You will need to add this to the beginning of the file:

```cpp
#include "../../../src/cs-utils/utils.hpp"
```
Can you confirm that this works?

michaelschleiss:
Works fine for me!

PR #333, opened 2023-02-23
fix contentType check

michaelschleiss:
Refers to #331.

### DIFF
--- a/plugins/csp-lod-bodies/src/TileSourceWebMapService.cpp
+++ b/plugins/csp-lod-bodies/src/TileSourceWebMapService.cpp
@@ -405,7 +405,9 @@ std::optional<std::string> TileSourceWebMapService::loadData(
     request.perform();
 
     auto contentType = curlpp::Info<CURLINFO_CONTENT_TYPE, std::string>::get(request);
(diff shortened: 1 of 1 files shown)


plugins/csp-lod-bodies/src/TileSourceWebMapService.cpp
```diff
@@ -404,7 +405,7 @@ std::optional<std::string> TileSourceWebMapService::loadData(
     request.perform();
 
     auto contentType = curlpp::Info<CURLINFO_CONTENT_TYPE, std::string>::get(request);
-    fail             = contentType != "image/png" && contentType != "image/tiff";
+    fail = !cs::utils::contains(contentType, "image/png") && !cs::utils::contains(contentType, "image/tiff");
```
Schneegans, line 408: I think clang-format would be happy if it was formatted like this:

```suggestion

    fail = !cs::utils::contains(contentType, "image/png") &&
           !cs::utils::contains(contentType, "image/tiff");
```

Schneegans (changes requested):
Awesome! Thanks a lot for directly fixing this. Could you try to commit the suggested change in order to make clang-format happy?

Schneegans (approved):

Schneegans:
Thanks again! We are looking forward to your future contributions :wink:

PR #335, opened 2023-03-06
:wrench: shader issue resolved

PdlAyush:
1. OpenGL error 1282.
2. Shader program with id 227,272,278 is not ready for use.
These errors existed when using `#version 330`.  The error was resolved by incrementing the versions of the atmospheres and simple-bodies shaders to `#version 430`.


JonasGilg:
Hi, thank you for your contribution. Can you please add a description to your pull request, describing, what the issue was that you solved and why you solved it like this?

I see, that you incremented the versions of the atmospheres and simple-bodies shaders. As far as I know they don't use any functionality of OpenGL 4.3.

PR #336, opened 2023-03-29
Increase maximum HEALPix depth

Schneegans:
With this, we can load much more high-resolution data. Changes:

* Increased maximum HEALPix depth from 20 to 30. This reduces our minimum tile size from about 13 m to 13 mm.
* Improved the scene scaling of the default configuration to allow for a smoother navigation close to the surface.
* Fixed an issue which led to an accumulated error in the rotation quaternion of the observer.
* Fixed an issue which caused precision issues for very small movements of the click-and-drag navigation.

With this set of changes, we can go well beyond a 1:1 scene scale and actually explore map data "like an ant" in the sub-millimeter range.

### DIFF
--- a/src/cs-core/DragNavigation.cpp
+++ b/src/cs-core/DragNavigation.cpp
@@ -137,14 +137,14 @@ void DragNavigation::update() {
   double const smoothThreshold = 0.8;
 
   if (mInputManager->pButtons[0].get() || mInputManager->pButtons[1].get()) {
-    glm::dvec3 end_vec;
-    glm::dvec3 start_vec;
+    glm::dvec3 endVec;
+    glm::dvec3 startVec;
     bool       bPerformRotation = false;
     mLocalRotation              = mInputManager->pButtons[1].get();
 
     if (mLocalRotation) {
-      start_vec        = mStartRayDir;
-      end_vec          = rayDir;
+      startVec         = mStartRayDir;
+      endVec           = rayDir;
       bPerformRotation = true;
     } else if (mDraggingPlanet) {
       // The radius is used to rotate the camera around the target body on a sphere of this exact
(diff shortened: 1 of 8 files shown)


src/cs-core/DragNavigation.cpp
```diff
@@ -154,15 +151,15 @@ void DragNavigation::update() {
...
-      auto t = IntersectSphere(rayOrigin, rayDir, rotationCenter, radius);
+      auto t = IntersectSphere(rayOrigin, rayDir, glm::dvec3(0.0), radius);
 
       // we do not want to drag something behind us
-      if (t && *t > 0) {
-        end_vec          = glm::normalize((rayOrigin + (*t * rayDir)) - rotationCenter);
-        start_vec        = glm::normalize(mStartIntersection - rotationCenter);
+      if (t && t.value() > 0.0) {
+        end_vec          = glm::normalize((rayOrigin + (t.value() * rayDir)));
+        start_vec        = glm::normalize(mStartIntersection);
```
JonasGilg, line 162: The two variable names do not conform to our guidelines.

src/cs-core/DragNavigation.cpp
```diff
@@ -182,9 +179,8 @@ void DragNavigation::update() {
         // The rotation axis is perpendicular to the start and the end position vectors
         mCurrentAxis = glm::normalize(currentAxis);
 
-        // The final amount of camera rotation around camlanetCenter
-        double dot         = glm::min(1.0, glm::dot(start_vec, end_vec));
-        double targetAngle = -1.0 * std::acos(dot);
+        // The final amount of camera rotation around planet center
```
JonasGilg, line 182:
```suggestion
        // The final amount of camera rotation around the body center
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #342, opened 2023-04-18
:wrench: Add error handling for TIFF functions

PdlAyush:
There are some scenarios where the error handling in (lines 69-76) is unable to detect and remove corrupt `.tiff` files in the `map-cache` directory. I was able to create one such scenario by:

- creating a spam request to spam the map-server
- using `std::this_thread::sleep_for ()` to create a delay before our ` curlpp::Easy request;` is performed.
While doing so I get the error:
```bash
TIFFFillStrip: Read error on strip `x`; got `y` bytes expected `z`.
```

The corrupt `.tiff` file wouldn't be removed. So the return of `TIFFGetField()` and `TIFFReadScanline()` is checked to remove such corrupt `.tiff` files.


plugins/csp-lod-bodies/src/TileSourceWebMapService.cpp
```diff
@@ -81,28 +81,39 @@ bool loadImpl(
...
-        std::vector<float> tmp(resolution);
-        TIFFReadScanline(data, tmp.data(), y);
-        int offset = resolution * y + (resolution - y);
-        int count  = y;
-
-        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
-        std::memcpy(
-            tile->data().data() + offset, tmp.data() + resolution - y, count * sizeof(float));
+    int tiffReturn{};
+    if (TIFFGetField(data, TIFFTAG_IMAGELENGTH, &imagelength)) {
```
Schneegans, line 85: I think it would be more readable if we would return early like in the case above. You will need to remove the `else` below and fix the indentation. So do not directly accept this but rather change this on your end :) 

```suggestion
    if (TIFFGetField(data, TIFFTAG_IMAGELENGTH, &imagelength) == 0) {
      logger().debug("TIFFGetField failed: Removing invalid cache file '{}'.", *cacheFile);
      boost::filesystem::remove(*cacheFile);
      return false;
    }
```

Schneegans (changes requested):
Thank you very much! I just added a cosmetic suggestion, else this looks good, I think.

PdlAyush:
Thank you for your feedback, sorry for the two commits, the latest commit is absolutely not needed as I am just using `TIFFGetField()`. Could you please only merge the [commit](https://github.com/cosmoscout/cosmoscout-vr/pull/342/commits/8bee368b587a87b6a0efd159ac11ff63b8df7b87).

Schneegans:
Thank you for implementing the changes. I think I cannot easily choose which commits to merge. Please undo the changes of that commit with a new commit. I'll will then squash-merghe this in the end.

Also, it would be great if you could add a note to the [changelog](https://github.com/cosmoscout/cosmoscout-vr/blob/main/docs/changelog.md) for this bug fix (under "Bug Fixes" for the upcoming v1.8.0).

Schneegans (approved):

PR #346, opened 2023-05-03
Show frame timing GUI in world-space per default

Schneegans:
This makes it possible to use the frame timings in an HMD.


plugins/csp-timings/src/Plugin.hpp
```diff
@@ -21,6 +22,12 @@ namespace csp::timings {
 /// timing statistics. This plugin can also be used to export recorded time series to CSV files.
 class Plugin : public cs::core::PluginBase {
  public:
+  struct Settings {
+    // If the statistics are shown on the local GUI area, the are drawn on each screen in a
```
JonasGilg, line 26:
```suggestion
    // If the statistics are shown on the local GUI area, they are drawn on each screen in a
```

JonasGilg (changes requested):

plugins/csp-timings/src/Plugin.hpp
```diff
@@ -21,6 +22,12 @@ namespace csp::timings {
...
+    // clustered setup.
```
JonasGilg, line 27:
```suggestion
    /// If the statistics are shown on the local GUI area, they are drawn on each screen in a
    /// clustered setup.
```

JonasGilg (changes requested):

JonasGilg (approved):

PR #347, opened 2023-05-04
Refactor LoD Bodies (continued)

Schneegans:
This PR continues the major refactoring of `csp-lod-bodies`. This time, about 1.5k lines of code have been removed. At the same time, the CPU performance has been greatly improved (a scene which required about 5ms on the CPU takes now about 1.5ms on my Linux workstation). Also, this PR paves the way to other improvements such as parallel tree traversal or multiple image channels. Here are the main changes:

* We now use RGBA instead of RGB image textures for better alignment which avoids costly uploads on some GPUs.
* The number of tiles which are uploaded per frame to the GPU has been reduced in order to prevent frame drops.
* More timing ranges are now displayed in order to show which parts of the planet rendering take the most time.
* A lot of indirections have been removed. The `TileNode` now simply stores a `TileData` for the elevation and image data. Before there were two quad trees managed in parallel, one for elevation tiles and one for image tiles. There were node searches and indirections via `TileId`s all over the place. This is now much more streamlined.
* A new struct `PerDataType` has been added. This is now used in places where were there used to be a `mMemberDEM` and a `mMemberIMG`. Consequently, a lot of duplicated code has been removed. Also, this will allow us theoretically to add more data channels in the future.
* The `UpdateBoundsVisitor` has been removed. Tile bounds are now recomputed only if required on-the-fly during `LODVisitor` traversal.


JonasGilg (changes requested):
Very nice, although on Windows I didn't see the same improvements in speed. Average GPU time went from 0.38 to 0.34 ms and average CPU time went from 0.51 to 0.31 ms.

JonasGilg (approved):

PR #348, opened 2023-05-20
Simple Bodies Ring Shadows

JonasGilg:
Simple bodies can now be shaded by their ring. They take the same config as the `csp-rings` plugin.

https://github.com/cosmoscout/cosmoscout-vr/assets/9581540/5f034843-47b0-4903-a982-b10da7634177

### DIFF
--- a/plugins/csp-simple-bodies/src/SimpleBody.cpp
+++ b/plugins/csp-simple-bodies/src/SimpleBody.cpp
@@ -12,7 +12,6 @@
 #include "../../../src/cs-graphics/TextureLoader.hpp"
 #include "../../../src/cs-utils/FrameStats.hpp"
 #include "../../../src/cs-utils/utils.hpp"
-#include "logger.hpp"
 
 #include <VistaKernel/GraphicsManager/VistaSceneGraph.h>
 #include <VistaKernel/GraphicsManager/VistaTransformNode.h>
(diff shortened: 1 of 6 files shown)


plugins/csp-simple-bodies/src/SimpleBody.cpp
```diff
@@ -152,9 +159,66 @@ float orenNayar(vec3 N, vec3 L, vec3 V) {
   
   return max(0, (L1 + L2) * cos_theta_i);
 }
-    
-void main()
-{
+
+// Calculates the shading by planetary rings. This is done in x steps:
```
JonasGilg, line 163:
```suggestion
// Calculates the shading by planetary rings. This is done in 3 steps:
```

plugins/csp-simple-bodies/src/SimpleBody.cpp
```diff
@@ -12,6 +12,7 @@
 #include "../../../src/cs-graphics/TextureLoader.hpp"
 #include "../../../src/cs-utils/FrameStats.hpp"
 #include "../../../src/cs-utils/utils.hpp"
+#include "logger.hpp"
```
Schneegans, line 15: I think that this is not required, or is it?

```suggestion
```

plugins/csp-simple-bodies/src/SimpleBody.cpp
```diff
@@ -152,9 +159,66 @@ float orenNayar(vec3 N, vec3 L, vec3 V) {
...
+
+    // Convert the distance to the texture coordinate.
+    float texPosition = (dist - uRingRadii.x) / (uRingRadii.y - uRingRadii.x);
+
+    // We apply the same shading logic as in the rings plugin.
+    #ifdef ENABLE_HDR
+      return (1.0 - texture(uRingTexture, vec2(texPosition, 0.5)).a) * 0.5;
+    #else
+      return (1.0 - texture(uRingTexture, vec2(texPosition, 0.5)).a) * 2.0;
+    #endif
```
Schneegans, line 215: I think this brightness variation is not required. In fact, it leads to weird artifacts especially in non-HDR mode. Also, in the original ring shader this is not applied to the alpha channel. I think this looks better:

```suggestion
    float texPosition = (dist - uRingRadii.x) / (uRingRadii.y - uRingRadii.x);
    return 1.0 - texture(uRingTexture, vec2(texPosition, 0.5)).a;
```

Schneegans (changes requested):
Thanks a lot! This is a simple but very nice addition. I added some remarks below.

Schneegans:
Looks good. However, I do not understand what Clang complains about. Do you have an idea?

Schneegans (approved):

PR #352, opened 2023-07-03
Update Vista and OpenSG

JonasGilg:
This PR updates Vista and OpenSG to the latest changes. This also allows us to skip the Window module in OpenSG and slightly improves build time.


Schneegans (approved):
Thanks a lot!

PR #360, opened 2023-09-06
Replace GLUT with SDL2

JonasGilg:
This PR replaces freeglut with SDL2 as a windowing toolkit, and input drivers. The main work is being done in Vista in the following PR: https://github.com/cosmoscout/vista/pull/24

One of the main features is a controller driver, which supports almost any controller that looks like a controller. Current feature support can be seen here: https://github.com/cosmoscout/vista/blob/feature/sdl-support/VistaCoreLibs/VistaDeviceDrivers/VistaSDL2ControllerDriver/README.md

I implemented a simple control scheme, that can be improved on in the future. Especially using the extended features of the PlayStation Controllers.


Schneegans:
Thank you very much for this impressive work! Everything seems to be working fine so far. However, when I run CosmoScout VR, I get these errors / warnings when loading the ini files:
*  **WARNING** No driver configurator for key [CONTROLLER_DB] in section of driver [SDL2CONTROLLER]
* ###ERROR### Unable to configure element type [DEFAULTWINDOW]
But everything seems to be working nonetheless. Can we get rid of these warnings?

docs/install.md
```diff
@@ -47,7 +47,7 @@ You should set these as required before executing the scripts below.
 This step only has to be done once.
 
 ```shell
-git submodule update --init
+git submodule update --init --recursive
```
Schneegans, line 50: There are at least two other places in the `docs` directory where the `--recursive` is missing.

Schneegans (changes requested):

Schneegans:
This looks very good to me. One last thing: Could you add a corresponding entry to the changelog?

JonasGilg:
> This looks very good to me. One last thing: Could you add a corresponding entry to the changelog?

Done.

Schneegans (approved):

PR #362, opened 2023-10-26
Node Editor Improvements

JonasGilg:
This PR adds two new features to the csl-node-editor and implements them in the csp-demo-node-editor.
1. `csl::nodeeditor::Node` now has a virtual `init()` function which can be implemented to do some initialization directly after the node is in a valid state.
2. Add a function to the `csl::nodeditor::NodeFactory` to register `Rete.Control` sublcasses, so they can be reused by multiple nodes.

### DIFF
--- a/plugins/csl-node-editor/src/NodeFactory.cpp
+++ b/plugins/csl-node-editor/src/NodeFactory.cpp
@@ -19,6 +19,12 @@ void NodeFactory::registerSocketType(std::string name, std::string color) {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+void NodeFactory::registerControlType(const std::string& controlSource) {
+  mControls.emplace_back(controlSource);
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
 std::string NodeFactory::getSocketSource() const {
   std::string source;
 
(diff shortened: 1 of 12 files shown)


plugins/csl-node-editor/src/NodeFactory.cpp
```diff
@@ -19,6 +19,12 @@ void NodeFactory::registerSocketType(std::string name, std::string color) {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+void NodeFactory::registerControlType(const std::string& controlSource) {
+  mControls.emplace_back(controlSource);
```
Schneegans, line 23: I think we usually pass things by value and then move them if they are to be copied.

```suggestion
void NodeFactory::registerControlType(std::string controlSource) {
  mControls.emplace_back(std::move(controlSource));
```

plugins/csl-node-editor/src/NodeFactory.hpp
```diff
@@ -43,6 +43,11 @@ class CSL_NODE_EDITOR_EXPORT NodeFactory {
     mNodeCreateFuncs[T::sName] = [=]() { return T::sCreate(args...); };
   }
 
+  /// Register a new control type. The string should contain a JavaScript class that inherits from
+  /// Rete.Control.
+  /// @param controlSource The source code of a JavaScript class derived from Rete.Control.
+  void registerControlType(std::string const& controlSource);
```
Schneegans, line 49: See above.

```suggestion
  void registerControlType(std::string controlSource);
```

Schneegans (changes requested):
Thank you very much. This looks very good!

Schneegans (approved):

PR #364, opened 2023-11-03
Add difference node to the node editor

NXXR:
Adds an Image2D Difference Node to the Node Editor under the "Operations" Category.
It has two Image2D Input sockets and calculates the difference between both.
Requires the bounds and size of the point vector (scalars) to be the same.


plugins/csp-visual-query/src/operation-nodes/DifferenceImage2D/DifferenceImage2D.js
```diff
@@ -0,0 +1,114 @@
...
+        switch (err) {
+        case 'DimensionMismatch':
+          elem.innerHTML += 'Input dimensions do not match';
+          break;
+        case 'BoundsMismatch':
+          elem.innerHTML += 'Input bounds do not match';
+          break;
+        case 'TimestampMismatch':
+          elem.innerHTML += 'Input timestamps do not match';
+          break;
```
JonasGilg, line 99: I think we should allow different timestamps.

JonasGilg (changes requested):

PR #370, opened 2024-02-02
Fix lod body loading

Schneegans:
This PR ensures that LoD bodies are properly re-registered as CelestialSurfaces after reloading the settings.


Flatken (approved):
Thanks for the fix. Looks ok now

PR #376, opened 2024-06-26
Enable vista text rendering

JonasGilg:
This fixes rendering bugs caused by text rendering in vista and adds a font to the vista config by default.


Schneegans (approved):
Thank you!

PR #379, opened 2024-07-16
Add fixed_horizon option to the ObserverNavigationNode

Schneegans:
This adds a `fixed_horizon` parameter to the ObserNavigationNode. If this is set to `true`, the observer is always rotated so that the horizon of the active object is always leveled.

For now, this breaks if we are in outer space or looking straight up or down. But it can already be very useful in cases were we know that the user is always close to a planet.

In the future, we could try making this work in outer space as well by checking the distance to the active object and gradually transitioning to the fixed horizon based on the distance.

You can enable this in DFN like this:

```xml
<node name="observer" type="ObserverNavigationNode">
   <param name="fixed_horizon" value="true" />
</node>
```

### DIFF
--- a/src/cosmoscout/ObserverNavigationNode.cpp
+++ b/src/cosmoscout/ObserverNavigationNode.cpp
@@ -9,8 +9,10 @@
 
 #include "../cs-core/SolarSystem.hpp"
 #include "../cs-gui/GuiItem.hpp"
+#include "../cs-utils/convert.hpp"
 
 #include <VistaAspects/VistaPropertyAwareable.h>
+#include <glm/gtx/io.hpp>
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


NXXR:
Could we use the ecliptic plane as the horizon in outer space?

Or would this just lead to weird rotations when entering or leaving a planet's frame?

Schneegans:
Yes, this could be possible. But a smooth transition is definitely non-trivial.

NXXR (changes requested):
Looks interesting.
Two Side effects I found to consider:
- When entering a body's frame the view is snapped to the horizon, since the observer enters the frame quite early (far from the the planet) the view can be snapped to the horizon quite rough, also leading to >90° instant rotations leading to disorientation.
- When traveling through outer space and clipping through a body's frame, it is possible that the observer is suddenly rotated even though this is unwantd (especially noticeable when approaching a body's frame by strafing up- or downwards).
- The adjustment of the rotation means that escaping the body's frame by strafing left or right is not possible anymore as this leads to a sideways orbit around the body. Strafing down leads to the observer getting (soft) stuck at the center of the body.

It should be made clear, that this option should only be used in use-cases where the observer is close to a planet's surface for the whole runtime, to avoid any of the mentioned issues.
This feature may also need to be revisited if it should be used in a more general context.

Schneegans:
Yes, that's why I wrote "this breaks if we are in outer space or looking straight up or down" :smile:.

NXXR:
> Yes, that's why I wrote "this breaks if we are in outer space or looking straight up or down" 😄.

Ah then I misunderstood.

I thought it meant:
> This breaks [the fixed horizon functionality] if we are in outer space.

My recommendation was more along the lines of:
> Do not use this when in outer space.
> This breaks [Cosmoscout navigation (and the users ability to not puke while using cosmoscout)] if we are [anywhere further away from a planet than a couple hundred thousand kilometers, but mostly] in outer space.

😁 I mean it works fine as a workaround and putting it into the main branch will do nothing if not enabled but I guess we should make a note or an issue somewhere to revisit this if it is supposed to be more than a workaround imho

NXXR (approved):

PR #383, opened 2024-10-18
Fix shadows on small moons

Schneegans:
The code which checked whether a body is potentially inside the shadow of another body had some flaws. This fixes that.


JonasGilg:
I think we still run into precision problems on moons like Mimas, but it is definetly an improvement to before.
image

JonasGilg (changes requested):
You are missing a changelog entry, other than that it looks good.

JonasGilg (approved):

PR #387, opened 2024-11-12
Reduce HDR flicker

Schneegans:
This pull request addresses a main issue of the HDR mode: As celestial objects get smaller on screen, they start to flicker. This is mainly because the discrete number of pixels they cover changes from frame to frame and therefore the luminance contribution to the framebuffer is temporally incoherent.

This PR fixes this to a certain degree by drawing a circular billboard at each objects position. These billboards have a minimum angular size, so they will always be visible regardless of the actual size of the object. As the real body gets smaller, the billboard is faded in to smoothly cover any flickering issues. As the real object gets smaller than its billboard, the luminance of the billboard is reduced to realistically simulate the apparent brightness of the object.

As a side effect, this allows us for the first time to render planets in the night sky with their proper luminance values.

Night Sky Before | Night Sky After
-------|------
image | image

## Before 
https://github.com/user-attachments/assets/072d8b16-b14e-4ee8-861b-7154156a49c6

## After
https://github.com/user-attachments/assets/c9bf4076-b4ed-4267-8a91-4187bed9db29

# Some Notes

* For the images and videos above I have disabled the planet marks. They are still available and enabled per default to show a bodies' position.
* This also changes how the solid angle of stars is computed. I think this is better. For small angles the difference is imperceptible, but the same formula is used for the billboards and here the difference was crucial.
* The configuration format of the `csp-trajectories` plugin is now a bit different. Most importantly, the color of the flare in HDR mode can be confgured (defaults to white) and should be set to approximately the albedo of the body (e.g. the average color of its texture).
* The `DeepSpaceDot` class of `csp-trajectories` is now much more versatile. It is now drawn in a perspective correct manner (not a circle but an ellipse in screen space) and used also for the non-HDR sun flare. Therefore the LDR sun flare looks a bit different now, but we do not need the `SunFlare` class anymore.


Schneegans:
So. I have just fixed a final issue. I think it's quite good now!

plugins/csp-trajectories/src/DeepSpaceDot.cpp
```diff
@@ -25,93 +25,91 @@ namespace csp::trajectories {
...
+
+#elif defined(HDR_BILLBOARD_MODE)
+  // This is basically a cone from above. The average value is 1.0.
+  float intensity = clamp(1.0 - dist, 0.0, 1.0) * 3.0;
+  oColor = vec4(uColor.rgb * intensity, uColor.a);
+
+#elif defined(FLARE_MODE)
+  // The quad is drawn ten times larger than the object it represents. Hence we should make the
+  // glow function equal to one at a distance of 0.1.
+  float alpha = 1.0 - pow(clamp((dist-0.1)/(1.0 - 0.1), 0.0, 1.0), 0.2);
```
JonasGilg, line 92:
```suggestion
  float alpha = 1.0 - pow(clamp((dist - 0.1) / (1.0 - 0.1), 0.0, 1.0), 0.2);
```

plugins/csp-trajectories/src/DeepSpaceDot.hpp
```diff
@@ -20,13 +19,47 @@
...
 
-  DeepSpaceDot(std::shared_ptr<Plugin::Settings> pluginSettings,
-      std::shared_ptr<cs::core::SolarSystem>     solarSystem);
+  /// When to draw the dot.
+  cs::utils::Property<int> pDrawOrder =
+      static_cast<int>(cs::utils::DrawOrder::eTransparentItems) - 1;
+
+  /// The color of the marker. In eHDRFlare mode, this corresponds to the average albedo of the
+  /// object.
+  cs::utils::Property<VistaColor> pColor = VistaColor(1, 1, 1, 1);
```
JonasGilg, line 53: Why are we using `VistaColor` here instead of a `glm::vec4`? We should use glm everywhere possible for consistency. It also would be easier to pass to the shader.

plugins/csp-trajectories/src/DeepSpaceDot.cpp
```diff
@@ -158,21 +176,24 @@ bool DeepSpaceDot::Do() {
...
   glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
   // draw simple dot
   mShader.Bind();
   glUniformMatrix4fv(mUniforms.modelViewMatrix, 1, GL_FALSE, glm::value_ptr(matMV));
   glUniformMatrix4fv(mUniforms.projectionMatrix, 1, GL_FALSE, glMatP.data());
-  mShader.SetUniform(mUniforms.color, pColor.get()[0], pColor.get()[1], pColor.get()[2]);
-  mShader.SetUniform(mUniforms.aspect, fAspect);
+  mShader.SetUniform(mUniforms.color, pLuminance.get() * pColor.get()[0],
+      pLuminance.get() * pColor.get()[1], pLuminance.get() * pColor.get()[2], pColor.get()[3]);
```
JonasGilg, line 191: If `pColor` would be a glm type it would be much more concise to pass it to the shader.

JonasGilg (changes requested):
The flickering is completely gone, nice idea!

Sadly there are two issues, one I find minor, but the other needs handeling:
1. (Minor issue): If you increase the focal length you can see the billboard instead of the body. This probably can be fixed, by fading it with the following ratio: `objectSolidAngle / viewportSolidAngle`.
2. (Major issue): If you are in an eclipse shadow and looking towards the Sun the user is seeing a bright object. This is because the billboards always have the max brightness, irregardless of the viewing angle. The user might think he sees the Sun, although the glare he is seeing is mistakenly from the occluders billboard. This could be fixed, by adjusting the brightness depending on the angle between the observer and the Sun.
**Edit**: If I am not mistaken you already calculate the phase angle and adjust brightness. I think the error might lie somewhere else, like the draw order?

Schneegans:
Thanks for the review. I'll look into it. I think the phase angle computation is mostly right (at least the angles, basing the brightness simply on the angles is a bit wonky). The flare turns dark if you look at it from the core shadow. However, auto-exposure makes it bright again as soon as you are a bit away from the core shadow.

But I think if we solve issue number one, the second issue will be solved as well since the flare will not be visible if you zoom in anymore.

Schneegans:
So, now the fading range depends on the field of view. I also made the size of the planet marks depend on the field of view, so that they are always about the same size on screen.

I also fixed another issue which made the black corners of the quad of the HDR flare around our Moon visible during daytime in the sky.

Schneegans:
Well, I'm not sure if we want to fix this with this PR. This is just our "usual" distance based culling. Simple bodies are not drawn if they are a hard-coded distance from the viewer. See here: https://github.com/cosmoscout/cosmoscout-vr/blob/main/src/cs-scene/CelestialObject.cpp#L187

JonasGilg:
> Well, I'm not sure if we want to fix this with this PR. This is just our "usual" distance based culling. Simple bodies are not drawn if they are a hard-coded distance from the viewer. See here: https://github.com/cosmoscout/cosmoscout-vr/blob/main/src/cs-scene/CelestialObject.cpp#L187

Maybe we can always draw the billboard, if the body has been culled away?

Schneegans:
Done. However, this is still not ideal because of the flare's maximum size limit, it will be smaller than the body if you zoom in a lot. If we increased the maximum size, it will hide the planet too early. So either we leave it as is, or we need to add a bit more intelligence here...

JonasGilg:
> Done. However, this is still not ideal because of the flare's maximum size limit, it will be smaller than the body if you zoom in a lot. If we increased the maximum size, it will hide the planet too early. So either we leave it as is, or we need to add a bit more intelligence here...

Okay, then we leave it as is for now. We can revisit it later.

JonasGilg (approved):

PR #391, opened 2024-12-14
Evaluate Bi-Cubic Glare-Interpolation in Compute-Preprocess for

Nyran:
This pull request implements a compute-shader-based bi-cubic glare-interpolation as a preprocessing step. Previously, the bi-cubic glare-interpolation was performed on-the-fly in the tonemapping pass which operates at window resolution granularity, although the glare-MipMaps are by default not computed at full resolution.

As a result of this pull-request, the combined processing time for computing glare-MipMap computation and Tonemapping pass in bi-cubic interpolation mode is reduced drastically (as an example: from ~1.3 ms down to ~0.7 ms using default Cosmoscout-VR settings on a Laptop equipped with an NVIDIA RTX 3070).

The first image shows an example of the old in-place bi-cubic interpolation performed **within the tonemapping pass**. The GPU execution time of the tonemapping pass (larg green bar in GPU timing overview) is much higher than the preprocessing time of the glare maps (violet bar below the green bar), which creates the data being consumed by the tonemapping pass consumes.
image

The second image shows an example of the new bi-cubic interpolation pre-processing approach, which reduces the tonemapping pass time drastically, while only slightly increasing the glare-map-computation time.
image

Future Work:
The compute-shader implementation is not optimal yet, since every thread in a compute-shader workgroup currently performs 16 independent texture lookups, despite high correlation between neighboring threads in a 2D pixel region. Follow-up work should investigate the effective use of shared memory in this scenario. First local implementation steps and tests indicate that the use of shared memory can reduce the combined run-time by another ~30%.

### DIFF
--- a/resources/shaders/glare.comp
+++ b/resources/shaders/glare.comp
@@ -27,17 +27,20 @@ const float PI = 3.14159265359;
 // the pixel position in the base layer of the output mipmap pyramid.
 // For performance reasons, we only use one sample for multisample inputs.
 vec3 sampleHDRBuffer(ivec2 pos) {
+  ivec2 pos_times_two = pos << 1;
+
 #if NUM_MULTISAMPLES > 0
-  vec3 col = imageLoad(uInHDRBuffer, ivec2(pos * 2 + ivec2(0, 0)), 0).rgb * 0.25 +
-             imageLoad(uInHDRBuffer, ivec2(pos * 2 + ivec2(1, 0)), 0).rgb * 0.25 +
-             imageLoad(uInHDRBuffer, ivec2(pos * 2 + ivec2(0, 1)), 0).rgb * 0.25 +
-             imageLoad(uInHDRBuffer, ivec2(pos * 2 + ivec2(1, 1)), 0).rgb * 0.25;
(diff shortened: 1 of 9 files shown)


Schneegans:
That's awesome! 

I wondered why the glare looks vertically squished and just realized that this was introduced in ffde74c773ff24cc94379da565d89276181cd6a1. I do not yet understand why that is. Also, I see some thin white bands:

This Branch | Before ffde74c773ff24cc94379da565d89276181cd6a1
-- | --
Screenshot from 2024-12-14 05-41-33 | Screenshot from 2024-12-14 05-49-21

But maybe we can merge this anyways and fix this later? Or do you want to have look at this right away?

Btw, we can ignore the failing checks until we try to merge this to `main`.

Nyran:
Thanks for putting the effort to find the visual difference! I did not have a good idea on how to recreate the exact same perspective and constellation across runs. I think it either way makes sense to expose the old mode as a third option both for debugging and exact performance analysis, so whoever touches the feature next could do this.

I do not mind whether we first merge the pull request or whether I'll try to fix it first. You can decide on what makes most sense from your perspective. I Assume that the squishing is an effect of the modified MipMal-Processing-Chain, such that potentialls the last vertical filter step is not applied or overriden?

PR #392, opened 2024-12-16
Make HDR flare and planet dot sizes configurable

Schneegans:


Schneegans:
@JonasGilg Shall we merge this also before the next version drops?

JonasGilg (approved):

Issue #393, opened 2024-12-17
Persistance of Window scaling and GUI activation when capturing a screenshot using the Web-API


DanielPSchenk:
## The Problem
When capturing a screenshot using the Web-API, the interactive window is resized and GUI state is set permanently. To me this is annoying when using the demo website and refreshing the view, I do not see a reason why web-API requests should change the interactive user interface for longer than necessary for taking the screenshot.

## The Solution
I added three variables to the Web-API plugin that keep restore values for viewport width, height and a bool for Gui activation state. Currently preparing a fork and feature branch for a pull request.

Schneegans:
Thanks for getting in touch. And that's a cool idea! Looking forward to the PR!

PR #394, opened 2024-12-17
🔧restore interactive window size and Gui state after taking Web-API c…

DanielPSchenk:
Extend the state of the Web-API plugin with variables for remembering the window size and Gui activation state before changes were made for a capture. Restore window size and Gui activation after taking a capture. Tested on Windows.
Implements Feature Request #393

### DIFF
--- a/plugins/csp-web-api/src/Plugin.cpp
+++ b/plugins/csp-web-api/src/Plugin.cpp
@@ -267,8 +267,9 @@ void Plugin::init() {
 
     // Read all paramters.
     mCaptureDelay  = std::clamp(getParam<int32_t>(conn, "delay", 50), 1, 200);
-    mCaptureWidth  = std::clamp(getParam<int32_t>(conn, "width", 0), 0, 2000);
-    mCaptureHeight = std::clamp(getParam<int32_t>(conn, "height", 0), 0, 2000);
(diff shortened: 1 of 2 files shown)


Schneegans:
Hey there, thanks a lot! This looks good.

The only issue I see that in some circumstances it may actually be bad to restore the state. Consider using CosmoScout VR as a virtual camera inside a hardware-in-the-loop simulation: Here, many images may be captured in short intervals. With this new feature, I would have to ensure that CosmoScout VR is started with the right resolution, else the window will resize itself all the time.

So maybe we could add an additional paramter to the `/capture` endpoint? Like for instance `restoreState=true`. Only when this is set, the window size and the gui will be restored after the capture. If you agree, you could add this about here: https://github.com/cosmoscout/cosmoscout-vr/blob/b2458e915bce5821d8d20900623a0ec1f93a10b3/plugins/csp-web-api/src/Plugin.cpp#L271

DanielPSchenk:
I added a restoreState parameter as suggested. On that occasion, I also increased the maximum resolution of web captures from 2000x2000 (too low for my use case) to 4096x4096, which works on my machine (Turing gen Nvidia laptop graphics)

Schneegans:
This looks good! Thank you very much!

Issue #397, opened 2025-01-07
HDRBuffer allowing access to color attachments outside of rendering code


DanielPSchenk:
## The Problem
The HDRBuffer uses raw VistaViewport pointers for keys of the unordered map. This results in errors when trying to get the color and depth attachments outside of rendering code, for example, in a Plugin update() function, because the function that is used for getting the current viewport returns a null pointer in such settings.
When I tried extending the web-api for capturing raw luminance values, I ran into this problem and it cost me ~5h of work realizing where the SegFaults were coming from, as there was no indication that all HDRBuffer functionality can only be used in rendering code.

## The Solution
I implemented a private method getCurrentViewportSafe that checks, if the viewport raw pointer is null, and if it is, it returns the key to the first element in the mHDRBufferData map. I changed getCurrentHDRBuffer to use this method so that it always returns valid HDRBufferData. 
This solution should interfere with no other functionality, unless there is some code that is built around getting an invalid HDRBufferData struct for some reason.
This solution can not be applied to the static getCurrentViewportSize and getCurrentViewPortPos functions because those have no access to the engine's mHDRBuffer.

I could make a PR soon after adding comments and stuff, together with the code for capturing raw luminance data over the web api.

Schneegans:
Hi again :smile: 

I think this sounds reasonable. I think it's pretty safe to assume that the buffer you get this way indeed contains the latest frame. Also, if I see it correctly, `getCurrentViewPortPos` is currently used nowhere and `getCurrentViewPortSize` only once. So if required, you could also refactor this part a bit.

DanielPSchenk:
okay, there will be a pull request soon. I will refactor `getCurrentViewPortPos` and `getCurrentViewPortSize`

PR #398, opened 2025-01-14
🔧implementing #397 🎉 Capturing raw luminance buffer contents.

DanielPSchenk:
Enabling the capture of raw luminance buffer contents over the web API. No image encoding involved, therefore it is faster on my machine than sending a png image, despite sending more data because of the higher quality and no compression. Very convenient to use with Python, just use numpy.frombuffer with float32 dtype and reshape to the image shape to get raw luminance values into Python.

Some behavior of the HDRBuffer became a roadblock when implementing this feature as discussed in #397. The code was refactored so that `GetVistaSystem()->GetDisplayManager()->GetCurrentRenderInfo()->m_pViewport;` is not used anymore as it returns a null pointer outside render threads. All functions previously using this functionality now use a new function that returns the pointer to the first viewport in the map if called outside a render thread. Therefore two functions can not be static anymore as they need to access the map but this is not a problem as there is no call to them where they are truly needed to be static.

### DIFF
--- a/plugins/csp-web-api/gui/example-web-frontend.html
+++ b/plugins/csp-web-api/gui/example-web-frontend.html
@@ -248,8 +248,8 @@ SPDX-License-Identifier: MIT
                 </tr>
                 <tr>
                   <td>format</td>
-                  <td>png <br> tiff <br> raw</td>
-                  <td>This can be either 'png', 'jpeg', or 'tiff'. It defaults to 'png' for color
+                  <td>png <br> jpeg <br> tiff <br> raw</td>
+                  <td>This can be either 'png', 'jpeg', 'tiff', or 'raw'. It defaults to 'png' for color
                     images and to 'tiff' for depth images. 'raw' transmits the contents of the luminance buffer without
                     color mapping.</td>
                 </tr>
--- a/plugins/csp-web-api/gui/example-web-frontend.html
+++ b/plugins/csp-web-api/gui/example-web-frontend.html
@@ -234,6 +234,13 @@ SPDX-License-Identifier: MIT
                   </td>
                 </tr>
                 <tr>
+                  <td>restoreState</td>
+                  <td>false</td>
+                  <td>If set to 'true', the window's size and the GUI visibility will be restored to
+                    the state before the capture.
+                  </td>
+                </tr>
+                <tr>
                   <td>delay</td>
                   <td>50</td>
                   <td>The application will wait this many frames before capturing the image.
--- a/plugins/csp-web-api/gui/example-web-frontend.html
+++ b/plugins/csp-web-api/gui/example-web-frontend.html
@@ -199,11 +199,13 @@ SPDX-License-Identifier: MIT
               --output image.png
             </div>
 
-            Images can be retrieved in the png, jpeg, and tiff formats.
+            Images can be retrieved in the png, jpeg, and tiff or raw formats. 
             Color images will always be returned with three 8 bits channels.
             Depth images will be returned as 32 bit floating point grayscale if the tiff format is
             used (which is default for depth image requests). If depth images are requested as png
             or jpeg file, they will be normalized to the 8 bit range.
+            Raw just transmits the contents of the luminance buffer without any color mapping as
+            32 bit floating point values.
 
             <table>
               <thead>
@@ -239,9 +241,10 @@ SPDX-License-Identifier: MIT
                 </tr>
                 <tr>
                   <td>format</td>
-                  <td>png <br> tiff</td>
+                  <td>png <br> tiff <br> raw</td>
                   <td>This can be either 'png', 'jpeg', or 'tiff'. It defaults to 'png' for color
-                    images and to 'tiff' for depth images.</td>
+                    images and to 'tiff' for depth images. 'raw' transmits the contents of the luminance buffer without
+                    color mapping.</td>
                 </tr>
                 <tr>
                   <td>depth</td>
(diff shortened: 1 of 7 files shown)


plugins/csp-web-api/gui/example-web-frontend.html
```diff
@@ -246,9 +248,10 @@
                 </tr>
                 <tr>
                   <td>format</td>
-                  <td>png <br> tiff</td>
+                  <td>png <br> tiff <br> raw</td>
                   <td>This can be either 'png', 'jpeg', or 'tiff'. It defaults to 'png' for color
```
Schneegans, line 252:
```suggestion
                  <td>png <br> jpeg <br> tiff <br> raw</td>
                  <td>This can be either 'png', 'jpeg', 'tiff', or 'raw'. It defaults to 'png' for color
```

plugins/csp-web-api/src/Plugin.cpp
```diff
@@ -466,18 +468,27 @@ void Plugin::update() {
...
-        } else if (mCaptureFormat == "png") {
-          stbi_write_png_to_func(&stbWriteToVector, &mCapture, mCaptureWidth, mCaptureHeight, 3,
-              capture.data(), mCaptureWidth * 3);
+        if (mCaptureFormat == "raw") {
+          // larger output size, but performance is comparable or better because there is no image encoding
+          mCapture.resize(mCaptureWidth * mCaptureHeight * 3 * 4);
+          std::shared_ptr<cs::graphics::HDRBuffer> hdrBuffer = mGraphicsEngine->getHDRBuffer();
+          VistaTexture* luminance_buffer = hdrBuffer->getCurrentWriteAttachment();
+          luminance_buffer->Bind(GL_TEXTURE_2D);
+          glGetTexImage(GL_TEXTURE_2D, 0, GL_RGB, GL_FLOAT, (void*)mCapture.data());
```
Schneegans, line 477: This produces `GL_INVALID_ENUM error generated. <texture> exceeds the maximum supported number of combined texture image units.` since it tries to bind the texture to the texture unit at 33984 :smile: 

```suggestion
          luminance_buffer->Bind();
          glGetTexImage(GL_TEXTURE_2D, 0, GL_RGB, GL_FLOAT, (void*)mCapture.data());
          luminance_buffer->Unbind();
```

Schneegans (changes requested):
Hi there again! Thanks a lot for this work! At a first glance, this looks good. I have only two  suggestion for improvement!
* What happens if HDR mode is currently disabled? I think it would be awesome if the 8-bit color from the framebuffer would be returned in this case. Just like the other formats, but in raw mode.
* The combination depth=true&format=raw is currently not supported. I think it would be a great addition if that worked as well!

What do you think?

DanielPSchenk:
thank you for pointing out the inconsistencies in the current implementation. It should be rather easy to also support raw depth buffers. I also agree that also supporting non-HDR would be useful. Not producing an error message would also be nice. I will work this out early next week.

DanielPSchenk:
I implemented the depth output reusing the tiff code, but instead of tiffWriteToVector using memcpy to write directly to mCapture (which was resized appropriately). @Schneegans looking forward to your feedback!

Schneegans:
Looks good so far! I think the Non-HDR-raw-color capture is still missing, right? You can check whether HDR Mode is currently enabled with `mAllSettings->mGraphics.pEnableHDR.get()`.

DanielPSchenk:
Non-HDR buffer output is implemented now. Raw output is in float so that code that uses this feature does not suddenly find a response a quarter its expected size.

plugins/csp-web-api/src/Plugin.cpp
```diff
@@ -466,18 +474,38 @@ void Plugin::update() {
...
+          // larger output size, but performance is comparable or better because there is no image
+          // encoding
+          mCapture.resize(mCaptureWidth * mCaptureHeight * 3 * sizeof(float));
+          if (mAllSettings->mGraphics.pEnableHDR.get()) {
+            // using the hdr buffer
+            std::shared_ptr<cs::graphics::HDRBuffer> hdrBuffer = mGraphicsEngine->getHDRBuffer();
+            VistaTexture* luminance_buffer = hdrBuffer->getCurrentWriteAttachment();
+            luminance_buffer->Bind();
+            glGetTexImage(
+                luminance_buffer->GetTarget(), 0, GL_RGB, GL_FLOAT, (void*)mCapture.data());
```
Schneegans, line 487: We usually try to avoid C-style casts.
```suggestion
            glGetTexImage(luminance_buffer->GetTarget(), 0, GL_RGB, GL_FLOAT,
                static_cast<void*>(mCapture.data()));
```

plugins/csp-web-api/src/Plugin.cpp
```diff
@@ -466,18 +474,38 @@ void Plugin::update() {
...
+            luminance_buffer->Unbind();
+          } else {
+            // without HDR, output is float in [0, 1], but the values in the buffer were
+            // previously converted to uint [0, 255]. For high quality raw output, use HDR mode.
+            glReadPixels(
+                0, 0, mCaptureWidth, mCaptureHeight, GL_RGB, GL_FLOAT, (void*)mCapture.data());
```
Schneegans, line 493: Same as above.
```suggestion
            glReadPixels(0, 0, mCaptureWidth, mCaptureHeight, GL_RGB, GL_FLOAT,
                static_cast<void*>(mCapture.data()));
```

src/cs-graphics/HDRBuffer.hpp
```diff
@@ -44,7 +44,7 @@ class CS_GRAPHICS_EXPORT HDRBuffer {
   HDRBuffer(HDRBuffer&& other)      = delete;
 
   HDRBuffer& operator=(HDRBuffer const& other) = delete;
-  HDRBuffer& operator=(HDRBuffer&& other) = delete;
+  HDRBuffer& operator=(HDRBuffer&& other)      = delete;
```
Schneegans, line 47: To make the clang-format version of the GitHub actions runner happy :smile: 
```suggestion
  HDRBuffer& operator=(HDRBuffer&& other) = delete;
```

Schneegans (changes requested):
Looks good! Just some very small things and then we are ready to go!

Schneegans (approved):

PR #399, opened 2025-01-21
🪲 remove the line through the pacific

DanielPSchenk:
Through the pacific ocean, there is a one-pixel wide line even if all GUI is disabled. The line only disappears when clouds are disabled.
with line
On closer inspection it appears there are not one, but 10 lines
line closeup
It is an issue with the sampling of the cloud texture at the line where u=1 and u=0 meet, similar to https://stackoverflow.com/questions/38357970/opengl-texture-repeat-artifacts.
I implemented the suggested solution of forcing a lower MAX_LOD, I found empirically, that at a maximum MipMap level of 5, the glitch reliably does not appear anymore.
no line
Before I figured this out, I debugged the cloud rendering. Is there a reason for using one-parameter atan(y_over_x) with logic for handling the different cases, instead of using two-parameter atan(y, x). I changed this to two-parameter atan and still got the same results, is there a reason I overlooked for using single-parameter atan?

### DIFF
--- a/plugins/csp-atmospheres/shaders/csp-atmosphere.frag
+++ b/plugins/csp-atmospheres/shaders/csp-atmosphere.frag
@@ -230,26 +230,7 @@ vec2 intersectOceansphere(vec3 rayOrigin, vec3 rayDir) {
 // centered at the origin.
 vec2 getLngLat(vec3 position) {
   vec2 result;
-
-  if (position.z != 0.0) {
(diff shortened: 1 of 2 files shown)


Schneegans:
Thanks for this fix! It looks much better this way! I guess that there is no particular reason for the one-parameter atan :sweat_smile:

Schneegans (approved):

PR #407, opened 2025-02-27
Less depth buffer copies

Schneegans:
The number of times we copied the depth buffer per frame has increased quite significantly. Once per atmosphere, once per sharad profile, once per WMS overlay. Soon I want to render some decals which will also require access to the depth buffer.

This PR adds a method to the `GraphicsEngine` which allows binding the current depth and color targets of the framebuffer as textures.

I have tested this with multiple viewports, but not in stereo.


Schneegans:
> I noticed that there is no changelog entry, but since there is no changes for the users, we might not need it.

Yeah, that was my reasoning, too.

PR #408, opened 2025-03-22
Improve performance of the luminance calculation

JonasGilg:
This PR improves the speed of the luminance calculation. Instead of doing a pyramid reduction it is using a more optimal reduction strategy. Depending on driver support there are two strategies:

1. Optimized parallel reduction (https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf)
Tries to make use of as many hardware optimizations as possible, especially burst memory access.

2. SubGroup based reduction (https://registry.khronos.org/OpenGL/extensions/KHR/KHR_shader_subgroup.txt)
On newer drivers GLSL has access to SubGroup support, assuring the most optimal ways to make use of each subgroup (aka: warp or wavefront).

In general I expected a speed up of 10x - 30x, based on NVIDIAs presentation, but only got a 1.5x speedup (tested on Windows only). Maybe modern drivers are able to optimize the original pyramid based reduction very well.

I originally wanted to improve the glare calculation as well, but since the performance improvement is so little, I put that off for now.


Schneegans (approved):
Thanks! Looks very good. But still interesting how little this speeds up the very basic initial implementation...

PR #409, opened 2025-04-08
3d clouds

DanielPSchenk:
Fully 3D clouds on a global scale. Built on ideas presented by Andrew Schneider at SIGGRAPH advances in real-time-graphics workshops. Adaped for cloud volumes the size of the earth.

The new clouds are fully 3D. To switch to the new cloud model toggle Advanced settings->Atmosphere->Advanced clouds.

Because ray marching is performed at a rather high resolution with secondary rays, the new system is really heavy on the GPU.

For designing cloudscapes, the new model uses additional textures. Defaults are provided and the configuration templates were modified to include those defaults. 

One error occurs when building, but it is not caused by the new code
`/cosmoscout-vr/src/cs-graphics/LuminanceMipMap.cpp(28,19): error C2065: "GL_SUBGROUP_SIZE_KHR" undeclared identifier`


Schneegans:
Thanks a ton! I'll look into it. Regarding the `GL_SUBGROUP_SIZE_KHR` error, you need to delete your `build/*-externals-Release/glew` directory and rebuild the externals. The glew version was updated recently.

Schneegans:
Thanks again! While not looking great for any vantage points, there are definitely perspectives where the clouds look pretty awesome. And on my 4070 they work quite decently. Before I look into the details, here are some high-level observations:
* Could you remove all the debug logging? The `read top texture at ...` even causes a crash if the texture is not defined
* I think the old clouds do not work if no `cloudTypeTexture` is set.
* Could you add some notes to the README what the `cloudTypeTexture` contains?
* I think it could be beneficial to move all the cloud shader code to a separate file at this point. What do you think? I mean it's just very few methods which are called by the `main`.

DanielPSchenk:
@Schneegans Are you changing the 2D or 3D noise or both?
About the points you brought up
1. sure, that should be removed
2. that's an oversight on my side, will fix this early next week
3. I can write a small documentation on what the texture does
4. that would be possible. GetCloudColor and everything it calls could be moved relatively easily, I will look into this possiblity next week
Good to hear that it works well on your machine

Schneegans:
> @Schneegans Are you changing the 2D or 3D noise or both?
> About the points you brought up

Neither, just the `uCloudTexture`. All I did was replacing
```glsl
float density = remap(textureLod(uCloudTexture, texCoords, 0).r, 0, CLOUD_COVER_MAX, 0, 1);
```
with
```glsl
float density = remap(textureLod(uCloudTexture, texCoords + vec2(uTime * 0.00002, 0.0), 0).r, 0, CLOUD_COVER_MAX, 0, 1);
```

DanielPSchenk:
I changed Atmosphere.cpp to only block the use of the default cloud model when no cloud texture is present. The advanced cloud model only works if both cloud texture and cloud type texture are present. There are now warnings when a cloud model is activated but the required texture was not specified.

DanielPSchenk:
I wrote a short documentation of what the different channels of the `cloudType` texture do

DanielPSchenk:
# Cloud System Rework V2

between_clouds

I again reworked he cloud system to now fully feature the procedural noise-based techniques from the Nubis cubed talked (SIGGRAPH 2022). Cloud prior density is now purely read from a global 2D texture and the cloud type is completely based on a 2D noise. This enables more complex multi-layered clouds than before.

Clouds are denser now, with the density being a function of height and cloud type. This allows for cirrus clouds and high altitude fog.
clouds_from_space
Anisotropic noises are still missing but I currently do not have the capcacity to implement them soon.

There are lots of quality of life improvements, for example all the most relevant parameters of the cloud model can be set through the GUI and WebAPI. 
clouds_from_space_more_high_clouds
For example the distribution of cloud types can be biased towards extremely tall clouds.

The cloud type texture is now easier to customize than before because it is automatically reloaded when the shader is compiled (causing a slight performance penalty but allowing for much faster iteration times). 
There are new global quality settings that are accessible through the GUI and WebAPI: a multiplier on all step sizes, the maximum number of steps taken during integration is exposed now and the sample jittering used to break up banding artifacts can be easily tuned.
clouds_with_mountains
The ray marching clouds blend naturally with the environment.

Performance is bad at high quality on a powerful laptop, but using the quality slider and by turning down sample jittering, stable (if banding-prone) results with good frame rates can be achieved.
performance
low quality, high performance
high_quality
high quality, low performance

An obvious next direction for performance would be half-resolution rendering, even very high quality runs at a stable 30fps+ on a high-end Ampere generation laptop if the resolution is only 960x560.

PR #410, opened 2025-04-13
Replace boost and fmt functionality with C++20 features and update CEF

JonasGilg:
This PR replaces:
- boost::posix_time with std::chrono
- boost::filesystem with std::filesystem
- fmt with std::format

This PR also updates CEF to version 135, to fix some C++ deprecation issues.

This reduces our dependencies and makes the setup of CosmoScout VR easier.

This change requires gcc-13 or Visual Studio 2019. gcc-13 is the default compiler on Ubuntu 24.04. There are some more possible improvements in date time parsing that would require gcc-14, but since this is not the default compiler on any LTS version of Ubuntu I skipped these improvements.


plugins/csp-recorder/README.md
```diff
@@ -53,5 +53,5 @@ Then run something like the following in CosmoScout's `bin` directory:
 3. **Encode the Frames:** Using something like `ffmpeg`, the individual frames can be merged to a video file.
 Here is an example:
    ```bash
-   ffmpeg -f image2 -framerate 60 -i frame_%d.png -c:v libx264 -preset veryslow  -qp 8 -pix_fmt yuv420p recording.mp4
+   ffmpeg -f image2 -framerate 60 -i frame_%d.png -c:v libx264 -preset veryslow  -qp 8 -pix_std yuv420p recording.mp4
```
Schneegans, line 56: This was not intended, I guess 😄 

```suggestion
   ffmpeg -f image2 -framerate 60 -i frame_%d.png -c:v libx264 -preset veryslow  -qp 8 -pix_fmt yuv420p recording.mp4
```

.github/workflows/deploy.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         sudo apt-get update -q
         sudo apt-get install libgl1-mesa-dev libglu1-mesa-dev libx11-dev
         sudo apt-get install libxi-dev libgconf-2-4
-        sudo apt-get install libboost-all-dev
+        sudo apt-get install
```
Schneegans, line 49: This can be removed, I guess.

```suggestion
```

Schneegans (changes requested):
Code-wise it looks good. I'll do some in-depth testing next.

JonasGilg:
> Code-wise it looks good. I'll do some in-depth testing next.

Thanks, this is not an urgent PR (also why this is in draft mode), since we currently have to support Ubuntu 22.04 which doesn't have gcc-13. Maybe in the next months we can drop Ubuntu 22.04 support. But you can review it nonetheless if you have the time.

Schneegans (approved):
I think everything looks good! Thanks for this work!

PR #412, opened 2025-04-24
Fix Visual Studio 2022 compilation

JonasGilg:
I couldn't compile CosmoScout with VS2022, due to multiple reasons. I updated OpenSG and spdlog and follow up errors in CosmoScout VR.


Schneegans:
This works for me. Is there anything missing?

Schneegans (approved):
Thanks!

PR #423, opened 2026-02-25
:green_heart: Fix windows externals build.

JonasGilg:
Fix a bug in the windows externals build, where it couldn't download bzip2. CMake can extract bzip files itself, so we just skip the download.


JonasGilg:
The linux build fails, because the coveralls website is down: https://status.coveralls.io/

Flatken (approved):
Thanks for fixing the issues with the windows build

PR #421, opened 2026-02-03
:beetle: Fix bug with atmosphere problems on one eye.

Slinyu:
Fix bug from issue #420 with atmosphere not rendering properly on one eye. This happened after #407 introduced the bug. The reason is that before #407 the depth and color buffers were copied on each Do(). Since #407 the copying through "mDirty = true" does not have the wished for effect. Now the "forceCopy" parameter is set to true and used to ensure copying properly.

PR #426, opened 2026-03-10
Fix error in sharad plugin that rendered wrong when HDR was disabled

JonasGilg:
Same fix as in #421.


Slinyu (approved):
Render works both in SDR and HDR.

PR #431, opened 2026-05-21
Fix linking bug in debug build for CEF on Windows.

JonasGilg:
On windows debug builds there is a mismatch in linker flags between CEF and CosmoScout (including all other dependencies).

CEF defines `_ITERATOR_DEBUG_LEVEL=0`, while the default in debug mode is `2`. This collides during linking.

My fix is that after the CEF download we modify the cef linker variables to fix this.

### DIFF
--- a/make_externals.bat
+++ b/make_externals.bat
@@ -378,6 +378,12 @@ IF NOT EXIST cef.tar.bz2 (
 
   rem We don't want the example applications.
   cmake -E remove_directory %CEF_DIR%/tests
+
+  rem Very ugly workaround for a linking bug, where CEF is build with different flags than the
+  rem rest of the project.
+  IF "%COSMOSCOUT_DEBUG_BUILD%"=="true" (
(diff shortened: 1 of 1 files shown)


jdfritsch (commented):
https://github.com/cosmoscout/cosmoscout-vr/blob/a7791300ac9e3bf54117691afc6d937c2b5c6616/src/cs-gui/gui.cpp#L63-L67
After checking gui.cpp, I don't think we actually use the CEF sandbox, so we could avoid this issue by just setting `-DUSE_SANDBOX=OFF` when building CEF.

JonasGilg:
> https://github.com/cosmoscout/cosmoscout-vr/blob/a7791300ac9e3bf54117691afc6d937c2b5c6616/src/cs-gui/gui.cpp#L63-L67
> 
> 
> After checking gui.cpp, I don't think we actually use the CEF sandbox, so we could avoid this issue by just setting `-DUSE_SANDBOX=OFF` when building CEF.

I see, isn't that a security risk? I think we currently do not restrict access to "external" websites.

jdfritsch:
> I see, isn't that a security risk? I think we currently do not restrict access to "external" websites.

Hm, fair point, although off the top of my head, I don't think there is much room for accessing arbitrary websites in the default config.
Still, this is probably something we should take a look at.

In that case, I think we can merge this PR and discuss the sandbox in another issue.

jdfritsch (approved):

PR #433, opened 2026-05-26
:beetle: Persist buffer object ids to prevent premature deletion

jdfritsch:
On a system running a RTX Pro 5000, the GLTF renderer encounters various issues when trying to display even a basic cube model.
This ranges from rendering only parts of a more complex model, over rendering seemingly nothing at all, to crashing as soon as the model gets into frame.

When investigating the issue, I found that the `shared_ptr`s responsible for managing and deleting data buffers of a `Primitive` go out of scope when the creation of the Primitive is finished and not when the `Primitive` itself goes out of scope.
This means, that all buffers belonging to `Primitive`s are in an undefined state, so it is kind of surprising that this worked on all other PCs so far.

This PR fixes this by simply storing a list of all allocated buffers as a member of `Primitive` objects.

### DIFF
--- a/src/cs-graphics/internal/gltfmodel.cpp
+++ b/src/cs-graphics/internal/gltfmodel.cpp
@@ -1154,6 +1154,13 @@ Primitive GltfShared::createMeshPrimitive(
     }
   }
 
+  // At this point, bufferMap contains the only reference to the OpenGL object ids
+  // for the newly created buffer objects, so unless we persist those in some way,
+  // the buffer memory may be cleared by the graphics driver.
+  for (auto const& bufferEntry : bufferMap) {
+    myPrimitive.bufferPtrs.push_back(bufferEntry.second.id);
+  }
+
   CheckGLErrors("createMeshPrimitive");
   return myPrimitive;
 }
--- a/src/cs-graphics/internal/gltfmodel.hpp
+++ b/src/cs-graphics/internal/gltfmodel.hpp
@@ -121,6 +121,7 @@ struct Primitive {
   glm::vec3 emissiveFactor{};
 
   std::vector<std::pair<Texture, TextureVar>> textures;
+  std::vector<std::shared_ptr<unsigned int>>  bufferPtrs;
   std::shared_ptr<unsigned int>               vaoPtr;
   std::shared_ptr<unsigned int>               programPtr;
 };


JonasGilg (approved):

JonasGilg:
Do you have an idea, why the pipeline checks were skipped?

JonasGilg:
> Do you have an idea, why the pipeline checks were skipped?

Ok, I downloaded the logs:

```
2026-05-26T10:26:51.2260000Z Evaluating comment_percentage.if
2026-05-26T10:26:51.2260000Z Evaluating: (success() && ((github.event_name == 'pull_request')))
2026-05-26T10:26:51.2260000Z Expanded: (true && ('push' == 'pull_request'))
2026-05-26T10:26:51.2260000Z Result: false
```

It looks like event_name is `push` instead of `pull_request`. Question is why now? Maybe GitHub had a change in the API?

jdfritsch:
> It looks like event_name is `push` instead of `pull_request`. Question is why now? Maybe GitHub had a change in the API?

Looking back at older PRs, they include both the checks for `push`, as well as for `pull_request` events.
So I think in this case the `pull_request` event didn't fire for some other reason.

Possibly related: I had some troubles submitting the PR, the first time I clicked "Create pull request" nothing happened except for the button turning inactive, so I reloaded the page a couple of times until it worked.
So maybe that broke some internal GitHub stuff...

JonasGilg:
> > It looks like event_name is `push` instead of `pull_request`. Question is why now? Maybe GitHub had a change in the API?
> 
> Looking back at older PRs, they include both the checks for `push`, as well as for `pull_request` events. So I think in this case the `pull_request` event didn't fire for some other reason.
> 
> Possibly related: I had some troubles submitting the PR, the first time I clicked "Create pull request" nothing happened except for the button turning inactive, so I reloaded the page a couple of times until it worked. So maybe that broke some internal GitHub stuff...

Since the change is so small, I don't see how it can cause a problem on any platform. If you agree I think we can merge it without the CI checks. If this happens again, we should investigate.

jdfritsch:
According to https://www.githubstatus.com/, GitHub is experiencing general problems with Actions. Maybe let's wait a while and see if the run is automatically triggered when the problems are fixed.

jdfritsch:
CI still hasn't run and there is a queued workflow run for this PR that is un-cancellable.

I say let's merge this and hope the CI works again for the next one...

PR #437, opened 2026-07-10
Main Menu

JonasGilg:
This branch introduces a new main menu system and migrates GUI native calls to a unified promise-based API, which allows the GUI to query information from C++.

New Features:
- Main Menu System: A new modal main menu accessible via a menu button in the toolbar
  - Plugin manager for loading, unloading, and reloading plugins at runtime
  - Scene save/load interface with a basic file browser
  - Exit to desktop option
- Promise Support for Native Calls:
  - JavaScript native calls now return Promises, enabling proper asynchronous operations
  - C++ callbacks can now return values that resolve to JavaScript Promises
- CosmoScout.callbacks API is now used consistently across all GUI components
- Updated Icon Font: Replaced MaterialIcons with MaterialSymbols Sharp
- Fixed: Timeline slider now uses positive-only values for time speed control


.github/workflows/build.yml
```diff
@@ -177,7 +177,7 @@ jobs:
           path: install/linux-Release/bin/test
 
   build_windows:
-    name: Windows MSVC 19.25
+    name: Windows MSVC 19.51
```
jdfritsch: Does it make sense to include the version number if we have (afaik) no way to set and pin the version of MSVC?

plugins/csl-tools/src/DeletableMark.cpp
```diff
@@ -7,6 +7,8 @@
 
 #include "DeletableMark.hpp"
 
+#include "logger.hpp"
```
jdfritsch: Not needed here as far as I see

plugins/csp-anchor-labels/gui/anchor_label.html
```diff
@@ -41,7 +41,13 @@
 </head>
 
 <body>
-  <div id="anchor-label" onclick="window.callNative('flyToBody')">
+  <script type="text/javascript" src="js/api.js"></script>
```
jdfritsch: Do we actually need to include `api.js` here?
Same for:
- `flag.html`
- `path.html`
- `polygon.html`
- `deletable_mark.html`

resources/gui/css/gui.css
jdfritsch, line 1: I'm not enough of a frontend dev to immediately spot where this comes from, but on a bright background you can notice the negative space of the rounded corners being slightly darker than the background.

Image

resources/gui/js/apis/statusbar.js
```diff
@@ -253,7 +253,8 @@ class StatusbarApi extends IApi {
               // completion.
               this._suggestionField.insertAdjacentHTML("beforeend", `<span class='${classNames}'
                        onclick='CosmoScout.statusbar._setCompletion(${prefixBegin}, ${prefixEnd}, 
-                                                              ${finalCursorPos}, "${completion}");'>
+                                                              ${finalCursorPos}, "${
```
jdfritsch: Kind of awkward linebreak, is this forced by clang format?

src/cosmoscout/Application.cpp
```diff
@@ -908,6 +911,10 @@ void Application::connectSlots() {
 void Application::registerGuiCallbacks() {
   // core callbacks --------------------------------------------------------------------------------
 
+  // Exits CosmoScout VR.
+  mGuiManager->getGui()->registerCallback(
```
jdfritsch, line 928: This crashes during shutdown in Visual Studio.
I think this is because `Quit()` shuts down the GUI, but after the callback function is executed, the WebView tries to resolve the corresponding promise.

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,498 @@
...
+
+    this._backButtonElement.addEventListener('click', () => this.back());
+
+    this._menuBodyElement.addEventListener('click', (event) => this._handleBodyClick(event));
+
+    this._menuElement.addEventListener('click', (event) => {
+      if (event.target === event.currentTarget) {
+        this.close();
+      }
+    });
```
jdfritsch: When clicking inside the menu, then dragging the cursor outside and releasing (or clicking outside and dragging inside), the menu closes.
Not sure what best UX practices are regarding this, but at least for me this felt slightly off.

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,498 @@
...
+   * Loads template content by ID.
+   * @param {string} id - The template ID to load.
+   * @returns {DocumentFragment|false} The template content or false if not found.
+   */
+  _loadTemplate(id) {
+    const template = CosmoScout.gui.loadTemplateContent(id);
+
+    if (template === false) {
+      console.warn(`Menu template '${id}' could not be loaded.`);
+    }
```
jdfritsch: Not sure if this error handling is meaningful here, `CosmoScout.gui.loadTemplateContent` already prints a warning.

resources/gui/js/apis/main_menu.js
jdfritsch: Should we also capture right clicks to close the menu?
Although, I just noticed that our UI generally does not capture right clicks, so rotating the scene is possible wherever you (right) click on the screen.

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,498 @@
...
+
+    this._loadSaveFiles().then((saveFiles) => {
+      const saveList = existingSavesSection.querySelector('.save-list');
+
+      if (saveFiles.length === 0) {
+        const emptyTemplate = this._loadTemplate('main-menu-save-empty-template');
+        if (emptyTemplate !== false) {
+          saveList.appendChild(emptyTemplate);
+        }
+        return;
```
jdfritsch: If you return here, the `save-new-button` is never assigned an event listener, so you are unable to ever create a scene config.

resources/gui/cosmoscout.html
```diff
@@ -631,6 +654,13 @@
           </a>
         </div>
         <div class="window-content">
+          <div class="row">
+            <div class="col-12">
+              <div class="calendar-wrapper">
+                <div id="calendar-container"></div>
+              </div>
+            </div>
+          </div>
```
jdfritsch: Does this actually do anything unless you also change the selectors in the initialization of the widget in `calendar.js`? https://github.com/cosmoscout/cosmoscout-vr/blob/e59005a3bbea2426c841f30da6cb35a4d313455e/resources/gui/js/apis/calendar.js#L32
https://github.com/cosmoscout/cosmoscout-vr/blob/e59005a3bbea2426c841f30da6cb35a4d313455e/resources/gui/js/apis/calendar.js#L42

src/cosmoscout/Application.cpp
```diff
@@ -1715,6 +1774,7 @@ void Application::registerGuiCallbacks() {
 void Application::unregisterGuiCallbacks() {
   mGuiManager->getGui()->unregisterCallback("core.save");
   mGuiManager->getGui()->unregisterCallback("core.load");
+  mGuiManager->getGui()->unregisterCallback("core.getSaveFiles");
```
jdfritsch, line 1788: `core.getPlugins` probably also needs to be unregistered.

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,498 @@
...
+
+  /**
+   * Handles overwriting an existing save file.
+   * @param {number} index - The index of the save file to overwrite.
+   */
+  _handleOverwriteSave(index) {
+    CosmoScout.callbacks.core.getSaveFiles().then((saveFiles) => {
+      if (index >= 0 && index < saveFiles.length) {
+        const name     = saveFiles[index].name;
+        const basename = name.split(/[\\/]/).pop();
```
jdfritsch: Since `getSaveFiles` already does truncate the files returned from cpp to their basename, this would be redundant, right?
Also, even if `getSaveFiles` could return paths containing slashes, shouldn't we pass these on in the `core.save` callback as is to make sure the right file is loaded?
(Same of course in `handleLoadScene`)

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,498 @@
...
+
+    CosmoScout.callbacks.core.save(filename)
+        .then(() => {
+          CosmoScout.notifications.print('Saved', `Scene saved as '${filename}'`, 'archive');
+          input.value = '';
+          this._renderSaveMenu();
+        })
+        .catch((err) => {
+          CosmoScout.notifications.print(
+              'Error', `Failed to save scene: ${err.message || err}`, 'error');
```
jdfritsch: Due to the async nature of this specific callback handler, I'm not sure if the promise is ever rejected.
E.g. if you try to save the scene to a non-existing directory, the UI will announce success, while on the console an error pops up.

jdfritsch (commented):
Some general observations:
* My first instinct was to try dragging the menu around, as it shares a lot of similarity with draggable windows. Maybe make the dimming effect of the background a bit more pronounced to signify this is something different?
* Sometimes after loading or unloading a plugin the list is not correctly updated for me, i.e. after unloading a plugin, the load and reload buttons are still shown. This is quite consistent when unloading `csp-anchor-labels` as a first step, but in other cases this can also happen rarely.
* Some plugins don't like being reloaded and crash instead. For csp-stars this is an easy fix (add `unregisterCallback` for `setDrawMode7`), but csp-lod-bodies has some inconsistent crashes involving its unit tests. Though I think fixing this could also be postponed.
* I was wondering if it would make sense to add all `.dll` / `.so` files in `share/plugins` to the list of available plugins, but I guess not all plugins have a sensible default config and expose all settings to the UI.

TODO (for me):
* [x] look at WebView

resources/gui/js/apis/statusbar.js
JonasGilg: Yes, this is clang-format ...

resources/gui/js/apis/main_menu.js
JonasGilg: Yeah, CosmoScout currently doesn't handle this. You can also interact through other UI items as well. Maybe this is a case for a different issue.

JonasGilg: I played around a little bit, but couldn't find an easy solution. I just added that it closes on mouse down outside the menu.

resources/gui/js/apis/main_menu.js
JonasGilg: > Since getSaveFiles already does truncate the files returned from cpp to their basename, this would be redundant, right?

`getSaveFiles` doesn't truncate the paths.

> Also, even if `getSaveFiles` could return paths containing slashes, shouldn't we pass these on in the `core.save` callback as is to make sure the right file is loaded?

I want to keep this as simple as possible. You can just load and save files in the bin folder for now. Especially the load function can only look in known places. So this is the easiest solution. Maybe later we can implement a default save location in the home directory (or AppData on Windows)?

I would also avoid touching the start config in `share/config` for now. This can serve as a fallback.

plugins/csp-anchor-labels/gui/anchor_label.html
JonasGilg: I removed most of them. Some need the api.

JonasGilg:
> * My first instinct was to try dragging the menu around, as it shares a lot of similarity with draggable windows. Maybe make the dimming effect of the background a bit more pronounced to signify this is something different?

There is now a full screen backdrop.

> * Sometimes after loading or unloading a plugin the list is not correctly updated for me, i.e. after unloading a plugin, the load and reload buttons are still shown. This is quite consistent when unloading `csp-anchor-labels` as a first step, but in other cases this can also happen rarely.

I fixed this by querying for actual change.

> * Some plugins don't like being reloaded and crash instead. For csp-stars this is an easy fix (add `unregisterCallback` for `setDrawMode7`), but csp-lod-bodies has some inconsistent crashes involving its unit tests. Though I think fixing this could also be postponed.

While my stars didn't cause a crash I added the missing unregister. The lod-bodies plugin should be investigated in another PR.

> * I was wondering if it would make sense to add all `.dll` / `.so` files in `share/plugins` to the list of available plugins, but I guess not all plugins have a sensible default config and expose all settings to the UI.

I thought so to, but I found it not feasible as a first step. Maybe you can also add the option to load individual configs for each plugin.

JonasGilg:
For the review you can skip the last two commits. They are just formatting. Somehow clang-format version 18.1.3 and 18.1.8 format stuff differently.

resources/gui/js/apis/main_menu.js
jdfritsch: > `getSaveFiles` doesn't truncate the paths.

Oops, mixed it up with `_loadSaveFiles` and thought the latter was used by the other functions. (Or misunderstood the `getSaveFiles` implementation.)

> I want to keep this as simple as possible. You can just load and save files in the bin folder for now. Especially the load function can only look in known places. So this is the easiest solution. Maybe later we can implement a default save location in the home directory (or AppData on Windows)?

I agree on keeping to the bin dir for now (although the save function actually allows any relative and absolute paths, as long as no new directories would have to be created (which I think is fine too)).

---

Still, I'm a bit torn on the trio of functions `_handleOverwriteSave`, `_handleLoadScene` and, to a lesser degree, `_loadSaveFile`. They work fine in the current setup, but:
* `_handleOverwriteSave` and `_handleLoadScene` receive some paths from cpp, do some destructive modifications to these, and then hand back the modified paths to cpp, hoping that they are equivalent. While this obviously works for all paths these functions can currently encounter, I think it would feel a bit cleaner to use the unmodified `name` in the callbacks `core.save/load`, and the truncated `basename` could be used only for the notification.
* When adding or removing scene files manually while on the load menu, clicking some name may load a completely different one, as the scene to load is selected by index and not name/id. Although I'm also fine labeling this as undefined behavior, as this requires fiddling with files in the install dir during runtime.
* Slightly inconsistent function names (_Scene_ vs _Save_)

jdfritsch:
> > * I was wondering if it would make sense to add all `.dll` / `.so` files in `share/plugins` to the list of available plugins, but I guess not all plugins have a sensible default config and expose all settings to the UI.
> 
> I thought so to, but I found it not feasible as a first step. Maybe you can also add the option to load individual configs for each plugin.

I also recently thought that a "layered" config system might be nice, where you can have our current configs as a base, but can then merge in additional configs with only minimal content (i.e. just a single plugin, or a single SPICE object added to multiple plugins).
This could actually be relatively easy to implement using JSON Merge Patches ([RFC 7386](https://tools.ietf.org/html/rfc7386)), which are already supported by nlohmann.
Only problem I see is handling adding items to bookmarks and downloadData, as arrays are overwritten during merge according to the RFC.

But this is definitely work for another PR.

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,497 @@
...
+      CosmoScout.notifications.print('Could not save!', 'Please enter a file name.', 'warning');
+      return;
+    }
+
+    if (!filename.endsWith('.json')) {
+      filename += '.json';
+    }
+
+    CosmoScout.callbacks.core.save(filename).then(() => {
+      CosmoScout.notifications.print('Saved', `Scene saved as '${filename}'`, 'archive');
```
jdfritsch: Maybe move this notification and the other equivalent ones in `_handleOverwriteSave`and `_handleLoadScene` over to `Application.cpp`, so that we can actually check for the result of the operation and display an appropriate notification for either success or failure.
https://github.com/cosmoscout/cosmoscout-vr/blob/4e8db1d77156f0b917b476a19d29103b8dde8372/src/cosmoscout/Application.cpp#L279-L294

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,497 @@
...
+    if (action === 'reload') {
+      CosmoScout.callbacks.core.reloadPlugin(pluginName);
+      callback = this._waitForPluginStateChange(pluginName);
+    } else if (action === 'unload') {
+      CosmoScout.callbacks.core.unloadPlugin(pluginName);
+      callback = this._waitForPluginStateChange(pluginName);
+    } else if (action === 'load') {
+      CosmoScout.callbacks.core.loadPlugin(pluginName);
+      callback = this._waitForPluginStateChange(pluginName);
+    }
```
jdfritsch, line 167: Nice, seems to update consistently now. Was just wondering, if this can lead to (much rarer) race conditions when a Plugin unloads fast enough to be already off by the time the initial check is done below?

JonasGilg:
> I also recently thought that a "layered" config system might be nice, where you can have our current configs as a base, but can then merge in additional configs with only minimal content (i.e. just a single plugin, or a single SPICE object added to multiple plugins). This could actually be relatively easy to implement using JSON Merge Patches ([RFC 7386](https://tools.ietf.org/html/rfc7386)), which are already supported by nlohmann. Only problem I see is handling adding items to bookmarks and downloadData, as arrays are overwritten during merge according to the RFC.
> 
> But this is definitely work for another PR.

I played around with JSON merge Patches in this branch a long time ago: https://github.com/cosmoscout/cosmoscout-vr/tree/feature/package-manager

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,497 @@
...
+        pluginItem.dataset.pluginName                                     = name;
+        pluginItem.querySelector('.plugin-manager-item-name').textContent = name;
+
+        pluginItem.querySelectorAll('[data-plugin-action]')
+            .forEach((actionButton) => { actionButton.dataset.pluginName = name; });
+
+        pluginList.appendChild(pluginItem);
+      });
+    });
+  }
```
jdfritsch, line 273: The clear of `_menuBodyElement.innerHTML` here and in `render()` lead to some annoying flickering on each action in the plugin list.

https://github.com/cosmoscout/cosmoscout-vr/blob/89585fcb678bc9c3355cabb7c72549a70cf24287/resources/gui/js/apis/main_menu.js#L511

A basic fix could be:
(Plain diff because suggestion marked all lines as changed)

```diff
  _renderPlugins() {
-   this._menuBodyElement.innerHTML = '';

    const pluginManager = CosmoScout.gui.loadTemplateContent('main-menu-plugin-template');
    if (pluginManager === false) {
      return;
    }

-   this._menuBodyElement.appendChild(pluginManager);

    CosmoScout.callbacks.core.getPlugins().then((plugins) => {
      if (this._currentMenu !== 'plugins') {
        return;
      }

      const pluginList     = pluginManager.querySelector('.plugin-manager-items');
      pluginList.innerHTML = '';

      Object.entries(plugins).forEach(([name, active]) => {
        const templateId = active ? 'main-menu-plugin-item-active-template'
                                  : 'main-menu-plugin-item-inactive-template';
        const pluginItem = CosmoScout.gui.loadTemplateContent(templateId);

        if (pluginItem === false) {
          return;
        }

        pluginItem.dataset.pluginName                                     = name;
        pluginItem.querySelector('.plugin-manager-item-name').textContent = name;

        pluginItem.querySelectorAll('[data-plugin-action]')
            .forEach((actionButton) => { actionButton.dataset.pluginName = name; });

        pluginList.appendChild(pluginItem);
      });
+     this._menuBodyElement.replaceChildren(pluginManager);
    });
  }
```

And also moving the clear in `render()´:

```diff
- this._menuBodyElement.innerHTML = '';

  if (menu.render) {
    menu.render();
    return;
  }

+ this._menuBodyElement.innerHTML = '';
```

Ideally the other `_render...` functions would have to be adapted too for consistency. However, it feels kind of awkward to force the `_render...` functions to take care of clearing the screen on their own. Any thoughts on this from your side?

src/cs-gui/WebView.hpp
jdfritsch:
```suggestion
  /// Calls an existing JavaScript function. You can pass as many arguments as you like. They will
```
Wennschon, dennschon ;)

src/cs-gui/WebView.hpp
```diff
@@ -64,7 +64,7 @@ class CS_GUI_EXPORT WebView {
     callJavascriptImpl(function, args);
   }
 
-  /// Execute Javascript code.
+  /// Execute JavaScript code.
   void executeJavascript(std::string const& code) const;
 
   /// Register a callback which can be called from Javascript with the
```
jdfritsch:
```suggestion
  /// Register a callback which can be called from JavaScript with the
```

Wennschon, dennschon ;)

src/cs-gui/WebView.hpp
```diff
@@ -234,26 +252,103 @@ class CS_GUI_EXPORT WebView {
...
+  /// @param promiseID The unique identifier of the Promise to resolve.
+  /// @param value     The value to resolve the Promise with (as a JSON string or JS literal).
+  void resolvePromise(double promiseID, std::string const& value) const;
+
+  /// Rejects a JavaScript Promise on the JavaScript side with the given error message.
+  /// This is used internally to reject Promises when a registered C++ callback throws an
+  /// exception or encounters an error.
+  ///
+  /// @param promiseID The unique identifier of the Promise to reject.
+  /// @param message   The error message to reject the Promise with (as a JSON string).
```
jdfritsch: If I understood the implementation of `rejectPromise` correctly, the `message` doesn't actually need to be a JSON string, but just any generic string, right?

jdfritsch (commented):
Haven't looked at the CMake/Vista changes yet, will do that over in that PR.

Btw, is it intended behavior that unloaded plugins are not persisted in saved scenes?
I.e., if I unload some plugin during runtime, save a new scene and then load the newly created scene, the plugin is loaded again.

resources/gui/js/apis/main_menu.js
JonasGilg: I tested it with the smalles plugins and it seems to work. I think the solution is already complicated enough.

resources/gui/js/apis/main_menu.js
JonasGilg: Now I seperated the info into name and path. The path contains everything, while the name only contains the file name without extension. Additionally I also show the Path in the save and load menus in a smaller font.

I tried to be more consistent with the save and scene words.

resources/gui/js/apis/main_menu.js
JonasGilg: Nice solution! I think it is fine for each function to clear the screen. We already have encapsulation by being inside the main_menu class.

src/cs-gui/WebView.hpp
JonasGilg: Now I renamed all wrongly spelled JavaScript occurences (also in function names 🙈).

JonasGilg:
> Btw, is it intended behavior that unloaded plugins are not persisted in saved scenes? I.e., if I unload some plugin during runtime, save a new scene and then load the newly created scene, the plugin is loaded again.

Nice catch. As it seems, this is a bug in the Settings class. It does persist also settings of unloaded plugins (so they can be reloaded) and it saves its state as is. Maybe a fix for this can be included in an overhaul of plugin/settings management. I would avoid touching it here, as this PR is already large enough.

src/cosmoscout/Application.cpp
```diff
@@ -945,11 +965,61 @@ void Application::registerGuiCallbacks() {
...
+        std::vector<std::map<std::string, std::string>> saveFiles{};
+
+        try {
+          auto saveDir = std::filesystem::current_path();
+          auto files = cs::utils::filesystem::listFiles(saveDir.string(), std::regex(".*\\.json$"));
+
+          for (auto const& file : files) {
+            std::map<std::string, std::string> fileInfo;
+            auto                               filePath = saveDir / file;
+            fileInfo["name"]                            = file;
```
jdfritsch:
```suggestion
            fileInfo["name"]                            = filePath.stem();
```

resources/gui/js/apis/main_menu.js
```diff
@@ -0,0 +1,486 @@
...
+   * @returns {Promise<Array<Object>>} Promise resolving to an array of scene file objects with
+   *     name, path, and date.
+   */
+  _loadSceneFiles() {
+    return CosmoScout.callbacks.core.getSceneFiles().then((sceneFiles) => {
+      return sceneFiles.map((file) => {
+        const name     = file.name || 'Untitled';
+        const basename = name.split(/[\\/]/).pop();
+        return {
+          name: basename.replace(/\.json$/i, ''),
```
jdfritsch:
```suggestion
        const name = file.name || 'Untitled';
        return {
          name: name,
```

src/cosmoscout/Application.cpp
jdfritsch: Upon further investigation, `listFiles` returns, unlike specified in its docstring, full paths (inconsistency was probably introduced during switch from boost to std::filesystem?). While that's also something we could fix down the line, I think we should just use the path's stem here to make this explicitly only the filename component.

resources/gui/js/apis/main_menu.js
jdfritsch: If we make sure that `core.getSceneFiles` actually returns only the basename as `file.name` (see other suggested change), we can remove the string manipulation here.

jdfritsch (changes requested):
I have two last suggestions/fixes for `getSceneFiles`/`_loadSceneFiles`.

Apart from that, this looks good to me!

jdfritsch (approved):
LGTM 👍

PR #439, opened 2026-07-20
Compatability with CMake 4+

### LINKED ISSUE #428: compile VS 18 2026 difficulties
Hi,

i try to compile on Win11 with Visual Studio 18 2026 -> i have problems with the modern cmake (since 3.5 they force minimum-version to 3.5) and boost (1.82.0 doesnt work standalone, and 1.90.0 from VS doesnt workl)

JonasGilg:
CMake 4+ removed support for features older than CMake 3.5. This branch fixes this by updating dependencies. The main update is in https://github.com/cosmoscout/vista/pull/29.

There are some caveats. Some dependencies, like SDL2_ttf or civetweb don't have a compatible branch and are stuck on old CMake versions. I added `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to their CMake build. This overrides whatever their build does internally with older features. It seems to be working for the few instances we had. It is not a nice solution.

This branch resolves #428.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,7 +153,16 @@ find_package(SDL2_ttf REQUIRED)
 find_package(OPENSG REQUIRED)
 find_package(CIVETWEB REQUIRED)
 find_package(VistaCoreLibs REQUIRED 
-  COMPONENTS "VistaBase" "VistaKernel" "VistaKernelOpenSGExt" "VistaOGLExt"
+  COMPONENTS
+    VistaAspects
+    VistaBase
+    VistaMath
+    VistaTools
+    VistaInterProcComm
+    VistaDataFlowNet
+    VistaKernel
+    VistaKernelOpenSGExt
+    VistaOGLExt
 )
 
 # X11 is used on Linux to set the application window's name and icon.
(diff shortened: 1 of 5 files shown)


jdfritsch:
Is there a reason for the inconsistencies between `make_externals.sh` and `make_externals.bat`?

I.e., on Linux, `CMAKE_POLICY_VERSION_MINIUM=3.5` is only added to civetweb, while on Windows it applies to civetweb, SDL2_ttf and curlpp.

JonasGilg:
> Is there a reason for the inconsistencies between `make_externals.sh` and `make_externals.bat`?
> 
> I.e., on Linux, `CMAKE_POLICY_VERSION_MINIUM=3.5` is only added to civetweb, while on Windows it applies to civetweb, SDL2_ttf and curlpp.

For SDL2_ttf the internal CMake logic used system libraries on Linux and self-vendored dependencies on Windows. The self-vendored libraries were not compatible with CMake 3.5+. To keep a consistent build I updated that the Linux build also uses self-vendored libraries.

curlpp just sets different CMake versions depending on the platform. Looks like it just sets the correct one on Linux.

jdfritsch (approved):
LGTM!

